PR #18, opened 2019-02-25
Use a loose version assembly loader for build task dependencies

JoeRobich:
We can help the assembly loader resolve build task dependencies by not being strict with version numbers. The Microsoft.Build.NuGetSdkResolver.dll depends on version 9 of Newtonsoft.Json, however the 3.0.100 preview 4 of the core sdk ships with version 10. The code for LooseVersionAssemblyLoader comes from [MSBuild CoreCLRAssemblyLoader](https://github.com/Microsoft/msbuild/blob/master/src/Shared/CoreCLRAssemblyLoader.cs#L137) with a change to load version of assemblies higher than what was requested.


JoeRobich:
@sharwell could you take a look?

JoeRobich:
> ❓ Why not supply an assembly binding redirect instead?

@sharwell  How would we know everything that needs redirection and what the appropriate version is for any given sdk? Is it possible to apply binding redirects from the sdk we are executing against?

jmarolf:
@JoeRobich 	you can also delete https://github.com/dotnet/format/blob/master/eng/Signing.props if you are no longer going to include Newtonsoft.Josn in the package

sharwell:
@JoeRobich What does MSBuild do in this case? Also, it sounded like the issue was Newtonsoft.Json, which means the full set of assemblies is just one assembly.

JoeRobich:
@sharwell  In other versions of the SDK it was the NuGet assemblies themselves. The requested version was 4.x and the version packaged in the SDK was 5.0. 

For `dotnet build` the binding redirects are baked into the 'dotnet.deps.json'

src/Program.cs
```diff
@@ -67,6 +67,13 @@ public static async Task<int> Run(string workspace, string verbosity, bool dryRu
                 // Since we are running as a dotnet tool we should be able to find an instance of
                 // MSBuild in a .NET Core SDK.
                 var msBuildInstance = Build.Locator.MSBuildLocator.QueryVisualStudioInstances().First();
+
+                // Since we do not inherit binding redirects from dotnet when referencing the SDK copy
```
nguerrera: "binding redirects" is a .NET Framework concept. I would say "since we do not inherit msbuild.deps.json."

nguerrera:
Since dotnet-format is its own process, there isn't much risk here.  So I approve this as a workaround, but this is basically an msbuild bug as we've discussed in the past, but haven't resolved.

I will review all of the tracking items we have for improving task loading on core and check that the >= issues in email were captured there.

@rainersigwald @vitek-karas PATAL as well.

PR #28, opened 2019-02-28
Since we have a 3.0.2 release version already we should update the pa…

jmarolf:
…ckage version

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <VersionPrefix>3.0.2</VersionPrefix>
+    <VersionPrefix>3.0.3</VersionPrefix>
     <PreReleaseVersionLabel>prerelease</PreReleaseVersionLabel>
   </PropertyGroup>
   <!--


eng/Versions.props
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <VersionPrefix>3.0.2</VersionPrefix>
+    <VersionPrefix>3.0.3</VersionPrefix>
```
sharwell: ❓ Can we not use GitVersioning to automate this? We should only need to change the version number for major and minor releases.

jmarolf: That's not how arcade works.  We need to add this feature. @tmat 	is there something I am missing? I would rather not use Nerdbank.Versioning and double our build time.

sharwell: 😕 I have not observed so substantial an impact with it.

/cc @AArnott

AArnott: I haven't seen it double build time either. [We have identified](https://github.com/AArnott/Nerdbank.GitVersioning/issues/114#issuecomment-428776274) that our assemblyinfo file generator contributes *most* of the time, which we can fix very easily. The GetBuildVersion task itself will be harder to optimize, but (depending on version commit depth) that represents perhaps only 25% of the Nb.gv addition to your build time after we fix the 75% problem.

jmarolf: Looking at telemetry, the 95% percentile for build time contribution for Nerdbank.Versioning was 12% of the build so my earlier statements where hyperbole. 12% build time regression is still significant in my opinion, but for this single project repo discussing a build time regression is a little ridiculous on my part. The real blocker is just that arcade does not have a feature like this yet right @tmat 	?

AArnott: I've pushed NB.GV v2.3.125 to nuget.org, which addresses a significant (perhaps 75%) source of the perf problem. I'd love to get your feedback on whether it is sufficient for your use.

jmarolf: Thanks @AArnott 	I'll also need to verify that it works with the unusual way global tool versioning is done but that is orthogonal to perf

AArnott: I'd love to hear the results of that verification, @jmarolf. I'm not sure what unusual way you're referring to. Several .NET global tools shipped on nuget.org are built with NB.GV today, FWIW.

jmarolf: Well it's more about _our_ weird way of doing it. We have pre-release and release versioning and they are different.  This doesn't play well with how global tools bake versioning into themselves. Perhaps this is all fully handled and I am worrying about nothing but it was the reason we unlisted the 3.0.0 package on nuget.org

tmat: Why do we need to bump the version? We only need to bump version when we release, like we do in Roslyn and other dotnet repos. Unique versions for non-release builds are automatically generated as documented in https://github.com/dotnet/arcade/blob/master/Documentation/CorePackages/Versioning.md. 
We are not gonna use NB.GV.

jmarolf: We already produced a 3.0.2 release build.  Each time we do that we will need to bump the version

tmat: I see. So we did release and thus need to bump the version. I think that's SOP.

tmat: Note that the revision number is used in both package version and assembly version, which are in sync to keep sanity. Changing assembly version every build makes referencing dependencies complicated (at least on Desktop FX). So we don't do that. That said, in some scenarios it might actually be beneficial to rev the assembly version. For example, for build task assemblies. For global tools it wouldn't probably matter that the assembly version changed each build. So we might do that as well. Perhaps we can add a flag to Arcade telling it to rev the version automatically.

AArnott: I'm not sure anyone is arguing for incrementing on every _build_. But incrementing for every commit is a thing that's becoming increasingly common.

FWIW, NB.GV supports incremental package versions more frequently than incrementing _assembly versions_. As @tmat says, changing the assembly version frequently isn't always the right choice. NB.GV defaults to setting major.minor as an assembly version, then setting all 4 integers into the assembly _file_ version and package version. But all these can be customized.

tmat: > But incrementing for every commit is a thing that's becoming increasingly common.

We have decided not to do that for .NET Core stack versioning.

PR #30, opened 2019-02-28
Set current directory to the workspace directory

JoeRobich:
After merging the change to use MSBuildLocator we no longer ensured the sdk set in the workspaces global.json would get picked up when locating msbuild. This change updates the current directory to the workspace's directory to restore this behavior.

### DIFF
--- a/src/Program.cs
+++ b/src/Program.cs
@@ -58,6 +58,8 @@ namespace Microsoft.CodeAnalysis.Tools.CodeFormatter
 
             try
             {
+                currentDirectory = Environment.CurrentDirectory;
+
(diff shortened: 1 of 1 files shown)


src/Program.cs
```diff
@@ -86,6 +89,13 @@ public static async Task<int> Run(string workspace, string verbosity, bool dryRu
             {
                 return 1;
             }
+            finally
+            {
+                if (!string.IsNullOrEmpty(currentDirectory))
+                {
+                    Environment.CurrentDirectory = currentDirectory;
```
JoeRobich: not sure if it is really necessary to restore the current directory when we are done.

PR #48, opened 2019-03-11
Add new CLI '--check' option

### LINKED ISSUE #41: Ability to check formatting / code style without writing to disk
Hello! 

Many existing code formatting tools ( [Prettier](https://prettier.io/), [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html) ) support checking code format without modifying source files. 

Currently to only check formatting I can implement some wrapper which will clone source code, run `dotnet-format' on this clone and then compare formatted files with original source code. But this seems to be too complex approach for such scenario. 

Do you plan to implement feature allowing to check format in `dotnet-format` (might be a console application option)?...

roman-petrov:
Closes #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Options:
                      the current directory for one.
   -v, --verbosity    Set the verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and
                      diag[nostic]
+  --check            Terminate with non-zero exit code if any files formatted.
   --version          Display version information
 ```
 
--- a/src/Program.cs
+++ b/src/Program.cs
@@ -32,13 +32,14 @@ namespace Microsoft.CodeAnalysis.Tools
                 .AddOption(new Option(new[] { "-w", "--workspace" }, Resources.The_solution_or_project_file_to_operate_on_If_a_file_is_not_specified_the_command_will_search_the_current_directory_for_one, new Argument<string>(() => null)))
                 .AddOption(new Option(new[] { "-v", "--verbosity" }, Resources.Set_the_verbosity_level_Allowed_values_are_quiet_minimal_normal_detailed_and_diagnostic, new Argument<string>() { Arity = ArgumentArity.ExactlyOne }.FromAmong(_verbosityLevels)))
                 .AddOption(new Option(new[] { "--dry-run" }, Resources.Format_files_but_do_not_save_changes_to_disk, new Argument<bool>()))
+                .AddOption(new Option(new[] { "--check" }, Resources.Terminate_with_non_zero_exit_code_if_any_files_formatted, new Argument<bool>()))
                 .UseVersionOption()
                 .Build();
 
             return await parser.InvokeAsync(args).ConfigureAwait(false);
         }
 
-        public static async Task<int> Run(string workspace, string verbosity, bool dryRun, IConsole console = null)
+        public static async Task<int> Run(string workspace, string verbosity, bool dryRun, bool check, IConsole console = null)
         {
             var serviceCollection = new ServiceCollection();
             var logLevel = GetLogLevel(verbosity);
--- a/src/Program.cs
+++ b/src/Program.cs
@@ -82,13 +82,14 @@ namespace Microsoft.CodeAnalysis.Tools
 
                 Build.Locator.MSBuildLocator.RegisterInstance(msBuildInstance);
 
-                return await CodeFormatter.FormatWorkspaceAsync(
+                var formatResult = await CodeFormatter.FormatWorkspaceAsync(
                     logger,
                     workspacePath,
                     isSolution,
                     logAllWorkspaceWarnings: logLevel == LogLevel.Trace,
                     saveFormattedFiles: !dryRun,
                     cancellationTokenSource.Token).ConfigureAwait(false);
+                return !check ? formatResult.ExitCode : (formatResult.FilesFormatted == 0 ? 0 : 1);
             }
             catch (FileNotFoundException fex)
             {
--- a/src/Program.cs
+++ b/src/Program.cs
@@ -89,7 +89,8 @@ namespace Microsoft.CodeAnalysis.Tools
                     logAllWorkspaceWarnings: logLevel == LogLevel.Trace,
                     saveFormattedFiles: !dryRun,
                     cancellationTokenSource.Token).ConfigureAwait(false);
-                return !check ? formatResult.ExitCode : (formatResult.FilesFormatted == 0 ? 0 : 1);
+
+                return GetExitCode(formatResult, check);
             }
             catch (FileNotFoundException fex)
             {
@@ -108,6 +109,9 @@ namespace Microsoft.CodeAnalysis.Tools
                 }
             }
         }
+
+        public static int GetExitCode(WorkspaceFormatResult formatResult, bool check) =>
+            !check ? formatResult.ExitCode : (formatResult.FilesFormatted == 0 ? 0 : 1);
 
         private static LogLevel GetLogLevel(string verbosity)
         {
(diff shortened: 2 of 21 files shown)


roman-petrov:
I just completed this PR and the implementation seems to work fine. I just added new `--check` option / flag. When this flag is specified `dotnet-format` CLI terminates with `1` exit code if any files in workspace formatted or should be formatted.

I updated existing unit tests according to my implementation. I used to following batch script to check if CLI application works as expected:

```bash
dotnet <dotnet-format.dll path> -v diag --check
if %ERRORLEVEL% GEQ 1 echo ERROR
if %ERRORLEVEL% EQU 0 echo OK
```

Now I am just removing `[WIP]` label and will wait for reviewer.

src/Program.cs
```diff
@@ -32,13 +32,14 @@ private static async Task<int> Main(string[] args)
...
                 .AddOption(new Option(new[] { "--dry-run" }, Resources.Format_files_but_do_not_save_changes_to_disk, new Argument<bool>()))
+                .AddOption(new Option(new[] { "--check" }, Resources.Terminate_with_non_zero_exit_code_if_any_files_formatted, new Argument<bool>()))
                 .UseVersionOption()
                 .Build();
 
             return await parser.InvokeAsync(args).ConfigureAwait(false);
         }
 
-        public static async Task<int> Run(string workspace, string verbosity, bool dryRun, IConsole console = null)
+        public static async Task<int> Run(string workspace, string verbosity, bool dryRun, bool check, IConsole console = null)
```
sharwell: 💡 Rather than pass this argument, we could _always_ simply return the number of files formatted from this method. The caller could then determine what to do with that number.

README.md
```diff
@@ -44,6 +44,7 @@ Options:
                      the current directory for one.
   -v, --verbosity    Set the verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and
                      diag[nostic]
+  --check            Terminate with non-zero exit code if any files formatted.
```
sharwell: ❔ Why not make this "the number of files formatted"?

roman-petrov: In current implementation if `--check` option specified the application exits with `1` exit code, I did this like specified in `Prettier` documentation <https://prettier.io/docs/en/cli.html#check>. 

Do like to change the implementation to make `exit code=<number of files formetted>` when `--check` option is specified?

sharwell: > Do like to change the implementation to make exit code=number of files formatted...?

I would not be a primary consumer of this feature, so other people with more experience using a `--check` option would need to answer your question. Your response will point those users to related documentation though which is a nice help. 😄

src/Program.cs
sharwell: 💭 Might need to use negative codes for errors, and positive codes for successful finding of number of files that need to be formatted. Otherwise the caller would not know whether 1 means error or no-error when `--check` is not given by the user.

jmarolf:
@roman-petrov looks good to me.  I'll wait for @JoeRobich 	's signoff but could you add a test for this please?

README.md
roman-petrov: So lets leave it as it is or change is still needed?:)

jmarolf: @roman-petrov 	I think the design that we doscussed for this is fine.

src/Program.cs
roman-petrov: > 💡 Rather than pass this argument, we could _always_ simply return the number of files formatted from this method. The caller could then determine what to do with that number.

Currently as I understand this method is being called by `CommandLineBuilder` command in code above:

`new Command("dotnet-format", 
handler: CommandHandler.Create(typeof(Program).GetMethod(nameof(Run))))`

New '--check' option is being passed in the same manner as all other options. So currently its not clear for me how to make the change you described, could you please advice?...

src/Program.cs
```diff
@@ -81,13 +82,14 @@ public static async Task<int> Run(string workspace, string verbosity, bool dryRu
 
                 Build.Locator.MSBuildLocator.RegisterInstance(msBuildInstance);
 
-                return await CodeFormatter.FormatWorkspaceAsync(
+                var formatResult = await CodeFormatter.FormatWorkspaceAsync(
                     logger,
                     workspacePath,
                     isSolution,
                     logAllWorkspaceWarnings: logLevel == LogLevel.Trace,
                     saveFormattedFiles: !dryRun,
```
JoeRobich: Does running a `--check` imply that `dryRun` should also be true or would you expect `--check` to update files on disk?

roman-petrov: The `--check` option will affect only program **exit code**. All other options, even if combined with `--check` options, will continue to work in the same manner as worked without my PR. I just tried to make minimal working implementation in my PR which will solve continuous integration scenarios.

roman-petrov:
> @roman-petrov looks good to me. I'll wait for @JoeRobich 's signoff but could you add a test for this please?
In my PR I modified existing unit tests by adding cased that check `formatResult.FileCount`, `formatResult.FilesFormatted`. 

To test new logic that program exit code equal to `1` when `--check` option is used I will have to write a unit test for `Program::Run` method, which takes 'IConsole' as last parameter. C# currently is not a primary language I work with, so can you please give an advice how to add good test for this?..

I can do some refactoring to test new logic but I wonder it might be too many changes.

JoeRobich:
@roman-petrov I think the easiest way to make that code testable would be to pull the logic that determines the final exit code into a static function. You could then test that logic in isolation by passing in different combinations of `check` and `formatResult`.

roman-petrov:
> @roman-petrov I think the easiest way to make that code testable would be to pull the logic that determines the final exit code into a static function. You could then test that logic in isolation by passing in different combinations of `check` and `formatResult`.

Thank you, I was thinking about the same :) I added some tests, please take a look.

src/Program.cs
roman-petrov: > 💭 Might need to use negative codes for errors, and positive codes for successful finding of number of files that need to be formatted. Otherwise the caller would not know whether 1 means error or no-error when `--check` is not given by the user.

Should we make some changes in this PR?..

tests/ProgramTests.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        [Fact]
+        public void ExitCodeIsZeroWithCheckAndNoFilesFormatted()
+        {
+            var formatResult = new WorkspaceFormatResult()
+            {
+                ExitCode = 42,
+                FilesFormatted = 0
+            };
+            var exitCode = Program.GetExitCode(formatResult, true);
```
JoeRobich: Use named arguments when passing boolean literals.
```suggestion
            var exitCode = Program.GetExitCode(formatResult, check: true);
```

tests/ProgramTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        [Fact]
+        public void ExitCodeIsSameWithoutCheck()
+        {
+            var formatResult = new WorkspaceFormatResult()
+            {
+                ExitCode = 42
+            };
+            var exitCode = Program.GetExitCode(formatResult, false);
```
JoeRobich: Use named arguments when passing boolean literals.
```suggestion
            var exitCode = Program.GetExitCode(formatResult, check: false);
```

src/Program.cs
JoeRobich: @roman-petrov I think the current implementation is probably what is expected. If we need to make a change, we can do it later.

roman-petrov:
Is there anything needed to change or this PR can be merged?

JoeRobich:
@roman-petrov No changes needed. Thanks for contributing!

roman-petrov:
@JoeRobich thanks all your help!

robinmanuelthiel:
Thanks @roman-petrov! Needed exactly this! <3

roman-petrov:
@robinmanuelthiel Welcome! 😃

JoeRobich:
@boukeversteegh Starting with net6.0 SDK dotnet-format ships with the CLI and there is no need to install it as a tool. Along with this change the `--check` option became `--verify-no-changes` to better align with the existing CLI options.

PR #51, opened 2019-03-13
update the language in the option description for --check

jmarolf:


README.md
```diff
@@ -44,8 +44,8 @@ Options:
                      the current directory for one.
   -v, --verbosity    Set the verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and
                      diag[nostic]
-  --check            Terminate with non-zero exit code if any files formatted.
   --dry-run          Format files, but do not save changes to disk.
+  --check            Terminate with non-zero exit code if any files need to be formatted in the workspace.
```
sharwell: 💡 It may be helpful to explicitly state that this option does not imply `--dry-run` (assuming that is still the case)

PR #52, opened 2019-03-14
Added files option

challen0:
I added a `--files` option that takes a comma-delimited string and only formats those files. I think I got everything, but if there's anything I missed or needs changing (or is wrong 😆), let me know!

This is for issue #9.


challen0:
@JoeRobich @jmarolf @sharwell I am getting build errors because an xlf file is out-of-date with the resx file. I'm not familiar with the xlf files. Is there something I need to do to generate them in the build script?

tests/CodeFormatterTests.cs
```diff
@@ -114,5 +114,26 @@ public async Task FSharpProjectsDoNotCreateException()
             Assert.Equal(0, formatResult.FilesFormatted);
             Assert.Equal(0, formatResult.FileCount);
         }
+
+       [Fact]
+        public async Task OnlyFormatFilesFromList()
```
JoeRobich: We need a `NoFilesFormattedWhenNotInList` test which would basically be this test, but with a non-existent file name in the files array

tests/CodeFormatterTests.cs
```diff
@@ -114,5 +114,26 @@ public async Task FSharpProjectsDoNotCreateException()
...
+        {
+            var logger = new TestLogger();
+            var path = Path.GetFullPath("tests/projects/for_code_formatter/unformatted_project_with_more_files/unformatted_project.csproj", SolutionPath);
```
JoeRobich: Instead of adding another project, let's add a new file to the unformatted project and update the tests.

src/CodeFormatter.cs
```diff
@@ -157,6 +157,23 @@ private static async Task<(Solution solution, int filesFormatted)> FormatFilesIn
                     continue;
                 }
 
+                if (filesToFormat != null)
+                {
+                    var fileInArgumentList = false;
```
JoeRobich: We should use a case insensitive comparison for the file names. Also, we would like to end the search after we've found a match.

```csharp
var fileInArgumentList = filesToFormat.Any(path => document.FilePath.EndsWith(path, StringComparison.OrdinalIgnoreCase));
```

challen0: What do you mean end the search when we find a match? Are you suggesting adding `break;` to the loop, which I guess `Any` takes care of in a sense?

tests/CodeFormatterTests.cs
challen0: Done!

tests/CodeFormatterTests.cs
challen0: Done!

src/CodeFormatter.cs
challen0: Done!

challen0:
@JoeRobich thanks for pointing out the broken XLF file!

I wanted to bring up an issue I ran into when testing. If you pass `path/to/file` (note forward slashes) on the CLI on Windows, the way the code currently works, it won't find it. Do you think that will be an issue?

JoeRobich:
@challen0 I think you could spin through the files calling `Path.GetRelativePath(Environment.CurrentDirectory, filePath)` to correct this for each platform. https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getrelativepath?view=netcore-2.2

zubivan:
Awesome, was also thinking to pick up this issue. Great work!

src/CodeFormatter.cs
```diff
@@ -160,7 +160,9 @@ private static async Task<(Solution solution, int filesFormatted)> FormatFilesIn
 
                 if (filesToFormat != null)
                 {
-                    var fileInArgumentList = filesToFormat.Any(path => document.FilePath.EndsWith(path, StringComparison.OrdinalIgnoreCase));
+                    var fileInArgumentList = filesToFormat
+                        .Select(path => Path.GetRelativePath(Environment.CurrentDirectory, path))
```
JoeRobich: Calling GetRelativePath here would have it fire over and over again for the same files for each document in the project. It would be better to move the GetRelativePath call to where we split the commandline argument so that we only run it once for each file.

challen0: Good point! Fixed it

src/Program.cs
```diff
@@ -142,5 +146,10 @@ private static void ConfigureServices(ServiceCollection serviceCollection, ICons
             serviceCollection.AddSingleton(new LoggerFactory().AddSimpleConsole(console, logLevel));
             serviceCollection.AddLogging();
         }
+        
+        private static string[] GetFileList(string files)
```
JoeRobich: If we make this method internal, it should be callable from the Test project. We could then test it for making directory separators uniform and for handling files that don't exist.

challen0: Good idea!

Issue #58, opened 2019-03-21
Don't know which files are formatted


challen0:
I noticed when running the tool, there's no way to tell which files were actually formatted just from reading the output. If I'm reading the code correctly, it prints 'Formatting file x' before it knows if changes will be made. It also looks like those messages are only available with verbosity of diagnostic. It would be nice to be able to run `--check` or `--dry-run` and know which files I need to address.

JoeRobich:
I think logging out the name of formatted files would be a good feature. Logging it as information is probably appropriate, that would have it printed when using the detailed verbosity.

JoeRobich:
This is now available on myget
```
dotnet tool install -g dotnet-format --version 3.0.4-prerelease.19174.1 --add-source https://dotnet.myget.org/F/format/api/v3/index.json
```

PR #61, opened 2019-03-22
Log formatted files

challen0:
This address #58 and prints a log message when a file is formatted. I put the message on the info level since the other messages about solutions/projects being formatted were that level too.

### DIFF
--- a/src/Resources.resx
+++ b/src/Resources.resx
@@ -190,6 +190,6 @@
     <value>The files to operate on. If none specified, all files in workspace will be operated on.</value>
   </data>
   <data name="Formatted_code_file_0" xml:space="preserve">
-    <value>Formatted '{0}'.</value>
+    <value>Formatted code file '{0}'.</value>
   </data>
 </root>
--- a/src/Resources.resx
+++ b/src/Resources.resx
@@ -189,4 +189,7 @@
   <data name="The_files_to_operate_on_If_none_specified_all_files_in_workspace_will_be_operated_on" xml:space="preserve">
     <value>The files to operate on. If none specified, all files in workspace will be operated on.</value>
   </data>
+  <data name="Formatted_code_file_0" xml:space="preserve">
+    <value>Formatted '{0}'.</value>
+  </data>
 </root>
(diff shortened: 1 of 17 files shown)


src/Resources.resx
```diff
@@ -189,4 +189,7 @@
   <data name="The_files_to_operate_on_If_none_specified_all_files_in_workspace_will_be_operated_on" xml:space="preserve">
     <value>The files to operate on. If none specified, all files in workspace will be operated on.</value>
   </data>
+  <data name="Formatted_code_file_0" xml:space="preserve">
+    <value>Formatted '{0}'.</value>
```
JoeRobich: This should be changed so that the name matches the value.

tests/CodeFormatterTests.cs
```diff
@@ -156,5 +156,22 @@ public async Task NoFilesFormattedWhenNotInList()
...
+        public async Task OnlyLogFormattedFiles()
+        {
+            var logger = new TestLogger();
+            var path = Path.GetFullPath("tests/projects/for_code_formatter/unformatted_solution/unformatted_solution.sln", SolutionPath);
+
+            var files = new[] {"Program.cs"};
+            var formatResult = await CodeFormatter.FormatWorkspaceAsync(logger, path, isSolution: true, logAllWorkspaceWarnings: false, saveFormattedFiles: false, filesToFormat: files, cancellationToken: CancellationToken.None);
+            var log = logger.GetLog();
+
+            var pattern = string.Format(Resources.Formatted_code_file_0, @"Program\.cs");
```
JoeRobich: I you formatted with the string `(.*)` instead, you could match with the pattern to extract the formatted file names out of the log.

src/Resources.resx
challen0: Thanks for pointing that out! I missed some text when I changed it.

tests/CodeFormatterTests.cs
challen0: I think I understand what you're saying, but I made a slight adjustment to the test to find the file name.

PR #88, opened 2019-04-08
Add integration test runner

JoeRobich:


JoeRobich:
@jmarolf please take a look

eng/format-verifier.ps1
```diff
@@ -0,0 +1,77 @@
...
+)
+
+$repos = @(
+    "https://github.com/dotnet/format",
+    "https://github.com/dotnet/roslyn"
+    "https://github.com/dotnet/cli",
+    "https://github.com/dotnet/project-system",
+    "https://github.com/Microsoft/msbuild",
+    "https://github.com/dotnet/corefx",
+    #"https://github.com/aspnet/AspNetCore",
```
jmarolf: Why not AspNetCore?

azure-pipelines-integration.yml
```diff
@@ -0,0 +1,25 @@
...
+- master
+
+# Branches that trigger builds on PR
+pr:
+- master
+
+jobs:
+
+- job: Windows
+  pool: dotnet-external-temp
```
jmarolf: I don't think we can use this pool anymore, it should be
```yaml
  pool:
    name: NetCorePublic-Int-Pool
    queue: buildpool.windows.10.amd64.vs2017.open
```

jmarolf:
You should probably more this pipeline into the dotnet/format folder in https://dev.azure.com/dnceng

eng/format-verifier.ps1
JoeRobich: It alone take more time than the uncommented repos combined.

azure-pipelines-integration.yml
```diff
@@ -0,0 +1,25 @@
...
+  strategy:
+    maxParallel: 4
+  timeoutInMinutes: 120
+  steps:
+    - script: eng\integration-test.cmd -logPath '$(Build.SourcesDirectory)\artifacts\log\integration' -testPath '$(Build.SourcesDirectory)\temp'
```
jmarolf: Why not define separate runs for each repository so they can happen in parallel?

```yaml
  strategy:
    maxParallel: 9
    matrix:
      format:
        _repo: "https://github.com/dotnet/format"
      roslyn:
        _repo: "https://github.com/dotnet/roslyn"
      cli:
        _repo: "https://github.com/dotnet/cli"
      project-system:
        _repo: "https://github.com/dotnet/project-system"
      msbuild:
        _repo: "https://github.com/Microsoft/msbuild"
      corefx:
        _repo: "https://github.com/dotnet/corefx"
      Blazor:
        _repo: "https://github.com/aspnet/Blazor"
      EntityFrameworkCore:
        _repo: "https://github.com/aspnet/EntityFrameworkCore"
      EntityFramework6:
        _repo: "https://github.com/aspnet/EntityFramework6"
```
you would just need to change the script to accept a repo URL
```yaml
- script: eng\integration-test.cmd -logPath '$(Build.SourcesDirectory)\artifacts\log\integration' -testPath '$(Build.SourcesDirectory)\temp' -repo $(_repo)
```

azure-pipelines-integration.yml
jmarolf: You can also run this on the big queue (the default one for AzDO) if we don't have any special VM dependencies  
```yaml
  pool:
    vmImage: 'vs2017-win2016'
```

jmarolf:
This looks good.  Just make sure that if `dotnet format` returns a non-zero exit code the build actually fails

jmarolf:
@JoeRobich you also might want to consider passing in a SHA to checkout.  Otherwise someone could break this repo by checking in elsewhere.  We would want to know about that happening, but I do not think we want to block unrelated PRs.

jmarolf:
:shipit:

Issue #102, opened 2019-04-18
dotnet-format doesn't find any files in legacy ASP.NET project


mus65:
dotnet-format doesn't format anything in an ASP .NET non-Core project because it doesn't find any files. I can even reproduce this with a newly created empty project in Visual Studio 2019.

```
# dotnet-format -v diag
Formatting code files in workspace '....\WebApplication6.csproj'
Loading workspace.
Workspace loaded in 1277ms.
Formatting code files in project 'WebApplication6'.
Formatted 0 of 0 files in 263ms.
Format complete.
```
I narrowed it down to the following line in the csproj file. If I remove it, everything works as expected:

`<Import Project="$(VSToolsPath)\WebApplications\Microsoft.WebApplication.targets" Condition="'$(VSToolsPath)' != ''" />
`

```
# dotnet-format -v diag
Formatting code files in workspace '....\WebApplication6.csproj'.
Loading workspace.
Workspace loaded in 2164ms.
Formatting code files in project 'WebApplication6'.
Formatting code file 'AssemblyInfo.cs'.
Formatting code file 'Class1.cs'.
Formatted code file 'Class1.cs'.
Formatted 1 of 3 files in 1535ms.
Format complete.
```

Version: 3.0.5-prerelease.19203.5+a138a5548eaa97c73d530260487fe8d3bec772a5

JoeRobich:
@mus65 Because dotnet-format runs against the .NET Core sdk copy of MSBuild, it doesn't have access to all of the build targets used by the various Full Framework project types. We might get lucky and be able to work around it without breaking other scenarios though. Thanks for narrowing it down!

JoeRobich:
As mentioned in https://github.com/dotnet/format/issues/102 using the `--folder` options instead of specifying a project or solution will format these files.

PR #131, opened 2019-05-02
Tweak MSBuild properties to allow enumerating ASP.NET full framework projects.

JoeRobich:
Clearing VSToolsPath will cause ASP.NET full framework projects to skip loading Microsoft.WebApplication.targets which does not exist in the .NET Core SDK.

Closes https://github.com/dotnet/format/issues/102

### DIFF
--- a/src/CodeFormatter.cs
+++ b/src/CodeFormatter.cs
@@ -114,6 +114,9 @@ namespace Microsoft.CodeAnalysis.Tools
                 // This flag is used at restore time to avoid imports from packages changing the inputs to restore,
                 // without this it is possible to get different results between the first and second restore.
                 { "ExcludeRestorePackageImports", bool.TrueString },
+                // Clearing VSToolsPath will cause ASP.NET full framework projects to skip loading Microsoft.WebApplication.targets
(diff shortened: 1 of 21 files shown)


src/CodeFormatter.cs
```diff
@@ -114,6 +114,9 @@ internal static class CodeFormatter
                 // This flag is used at restore time to avoid imports from packages changing the inputs to restore,
                 // without this it is possible to get different results between the first and second restore.
                 { "ExcludeRestorePackageImports", bool.TrueString },
+                // Clearing VSToolsPath will cause ASP.NET full framework projects to skip loading Microsoft.WebApplication.targets
+                // which does not exist in the .NET Core SDK.
+                { "VSToolsPath", string.Empty }
```
JoeRobich: @jmarolf Not sure what else might fallout because of this change

jmarolf: I would argue the correct fix is to do what Omnisharp does and deploy its own MSBuild that can load legacy tasks and targets

JoeRobich:
I think if we were to pursue a fix for this, it would be to allow users to specify build properties. Closing

Issue #111, opened 2019-04-20
EOF newline not added


supergibbs:
I have `insert_final_newline = true` in my `.editorconfig` but `dotnet-format` doesn't insert it.

FWIW, VS Code inserts it no matter what `insert_final_newline` is set to.

sharwell:
@JoeRobich Can you set the resolution and milestone for this?

PR #133, opened 2019-05-03
Added final newline formatter

JoeRobich:
Adds a new formatter for the `insert_final_newline` editorconfig option.

Closes https://github.com/dotnet/format/issues/111


src/Formatters/EOFNewlineFormatter.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Options;
+using Microsoft.CodeAnalysis.Text;
+using Microsoft.Extensions.Logging;
+using Microsoft.VisualStudio.CodingConventions;
+
+namespace Microsoft.CodeAnalysis.Tools.Formatters
+{
+    internal sealed class EOFNewLineFormatter : DocumentFormatter
```
sharwell:
```suggestion
    internal sealed class EndOfFileNewLineFormatter : DocumentFormatter
```

JoeRobich:
@sharwell I had assumed that we were letting the Editor handle some of these core .editorconfig options.  I can add it there instead.

sharwell:
Not all options would apply there, but it seems these two would.

JoeRobich:
@sharwell I see that Roslyn already has an issue to support final newline in the formatter (https://github.com/dotnet/roslyn/issues/33468). Going to merge this and we can remove when the functionality is implemented in Roslyn.

Issue #247, opened 2019-07-10
Cannot write problem matcher for VS Code custom task


AnthonyMastrean:
I'm trying to run `dotnet format` in a VS Code custom task. But, there's not enough detail in the output to write a decent [problem matcher](https://code.visualstudio.com/docs/editor/tasks#_defining-a-problem-matcher).

**tasks.json**
```typescript
"problemMatcher": {
    "fileLocation": ["relative", "${workspaceFolder}"],
    "pattern": {
        "regexp": "^\\s*Formatted code file '(.*)'.$",
        "file": 1,
        "line": 2,
        "column": 3,
        "severity": 4,
        "message": 5
    }
}
```

The output doesn't provide enough of a file path for `fileLocation`. It doesn't provide line or column numbers. And it doesn't provide a specific message.

```
$ dotnet format --check --dry-run
  Formatting code files in workspace 'C:\Users\anthony\example\src\Example\Example.csproj'.
  Formatting code files in project 'Example'.
  Formatted code file 'Startup.cs'.
  Format complete.
```

# Considerations

The important points to consider when redesigning the output:

 * The output should include _workspace relative_ paths, no matter how I specify the `--workspace`
 * The output should include line and column number where the violation occurs
 * The output should not include severity, since that's not a relevant concept; the user can choose whichever is appropriate for their project: `warning` or `error`
 * The output should include not just the [rule name](https://github.com/dotnet/format/wiki/Supported-.editorconfig-options), but whether the rule was required or not and how it was violated (more thought on this is probably required.. maybe append "not required" for `false` rules and "required" for `true` rules?)

## Simple Problem Matcher
https://code.visualstudio.com/Docs/editor/tasks#_defining-a-problem-matcher

**template**
```
[workspace relative file path](line,col): [rule name]
```

**output**
```
$ dotnet format --check --dry-run
Formatting code files in workspace 'C:\Users\anthony\example\Example.sln'.
  src\Example\Startup.cs(26,22): space between parentheses not required
  src\Example\Controllers\ExampleController.cs(100,31): new line before open brace required
Format complete.
```

**tasks.json** ([regex101](https://regex101.com/r/ewOwqy/1))
```typescript
{
    "label": "format",
    "command": "dotnet",
    "type": "process",
    "args": ["format", "--check", "--dry-run"],
    "problemMatcher": {
        "fileLocation": ["relative", "${workspaceFolder}"],
        "severity": "warning",
        "source": "dotnet format",
        "pattern": {
[...]

PR #249, opened 2019-07-12
Log formatting change details when in dry-run mode

### LINKED ISSUE #90: [Feature Request] Detailed logging to include line number and code violation details
I'd like to be able to run this tool on a codebase and have it give me a list of the files + associated line numbers and the rule violation, we'd use this information to plumb back into our CI reports.

At the moment all we know is that a particular file needed modifying, nothing more specific.

It looks like the changes are applied at a document level currently so I'm not sure if this level of fidelity is possible. Presumably as VS can offer it up however, then there must be a way.

JoeRobich:
Provides the necessary information for #247. Since the Roslyn whitespace formatter fixes all whitespace formatting options in one operation there are not per-rule messages for whitespace changes.

Sample output:
```
~\source\repos\format [dev/jorobich/log-format-locations ≡ +0 ~8 -0 !]> dotnet format --dry-run
  Warnings were encountered while loading the workspace. Set the verbosity option to the 'diagnostic' level to log warnings.
  src\Formatters\DocumentFormatter.cs(30,9): Fix whitespace formatting.
  Formatted code file 'DocumentFormatter.cs'.
  Format complete in 5337ms.
```

Fixes #90

### DIFF
--- a/src/Formatters/FinalNewlineFormatter.cs
+++ b/src/Formatters/FinalNewlineFormatter.cs
@@ -13,7 +13,7 @@ namespace Microsoft.CodeAnalysis.Tools.Formatters
 {
     internal sealed class FinalNewlineFormatter : DocumentFormatter
     {
-        protected override string FormatWarningDescription => "Add final newline.";
+        protected override string FormatWarningDescription => Resources.Add_final_newline;
 
         protected override async Task<SourceText> FormatFileAsync(
             Document document,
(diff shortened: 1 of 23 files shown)


JoeRobich:
@jmarolf PTAL

src/Formatters/FinalNewlineFormatter.cs
```diff
@@ -13,10 +13,13 @@ namespace Microsoft.CodeAnalysis.Tools.Formatters
 {
     internal sealed class FinalNewlineFormatter : DocumentFormatter
     {
+        protected override string FormatWarningDescription => "Add final newline.";
```
jmarolf: do we want to add localization for this stuff?

PR #268, opened 2019-07-23
Run analyzers initial impl

jmarolf:
making anaysis happen in parallel

### DIFF
--- a/src/Analyzers/CodeAnalysisResult.cs
+++ b/src/Analyzers/CodeAnalysisResult.cs
@@ -1,11 +1,40 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
 
-using System;
+#nullable enable
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+
+using NonBlocking;
 
 namespace Microsoft.CodeAnalysis.Tools.Analyzers
 {
     public class CodeAnalysisResult
     {
-        internal void AddDiagnostic(Document document, Diagnostic diagnostic) => throw new NotImplementedException();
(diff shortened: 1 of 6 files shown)


src/Analyzers/CodeAnalysisResult.cs
```diff
@@ -1,11 +1,40 @@
...
+                     return list;
+                 },
+                updateValueFactory: (key, list) =>
+                {
+                    list.Add(diagnostic);
+                    return list;
+                });
+        }
+
+        public IReadOnlyDictionary<DocumentId, ImmutableArray<Diagnostic>> Diagnostics
```
RikkiGibson: I need this to proceed with PerDocumentCodeFixApplier

src/Analyzers/ConcurrentAnalyzerRunner.cs
```diff
@@ -21,38 +20,32 @@ internal class ConcurrentAnalyzerRunner : IAnalyzerRunner
...
-
-            var documents = formattableDocuments.Select(x => x.Document).ToList();
             var result = new CodeAnalysisResult();
-            foreach (var project in solution.Projects)
+            var solution = formattableDocuments.FirstOrDefault().Document.Project.Solution;
+            var documents = formattableDocuments.Select(x => x.Document).ToList();
+            Parallel.ForEach(solution.Projects, project =>
             {
-                var compilation = await project.GetCompilationAsync(cancellationToken);
+                var compilation = project.GetCompilationAsync(cancellationToken).GetAwaiter().GetResult();
```
JoeRobich: Should we check if the project is a C# or VB project?

src/Analyzers/CodeAnalysisResult.cs
jmarolf: yep.  I need to fix integration tests first.  You can assume that this is the shape of the object you will be given

PR #276, opened 2019-07-23
Update to ExternalAccess.Format with LegacyEditorConfig provider

JoeRobich:


src/CodeFormatter.cs
```diff
@@ -6,8 +6,10 @@
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Reflection;
```
RikkiGibson: is this used?

src/Utilities/FormatHostServices.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    private static MefHostServices s_hostServices;
+    public static MefHostServices HostServices
+    {
+        get
+        {
+            if (s_hostServices == null)
+            {
+                var host = MefHostServices.Create(FormatHostServices.DefaultAssemblies);
+                Interlocked.CompareExchange(ref s_hostServices, host, null);
```
jmarolf: probably not necessary

PR #282, opened 2019-07-24
Clean up

jmarolf:
just some moving things around. not super important


jmarolf:
> What do the NuGet.config entries do?

eventually we will need these changes for arcade.  I am just being proactive.  As a binus Nuget restore should be more correct now.

jmarolf:
Actually nevermind, the arcade changes that will force this on us haven't been merged yet

PR #283, opened 2019-07-25
Enable other Frameworks

deinok:
This change should enable other frameworks to work

### DIFF
--- a/src/dotnet-format.csproj
+++ b/src/dotnet-format.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp2.2;netcoreapp3.0</TargetFrameworks>
     <OutputType>Exe</OutputType>
     <LangVersion>7.3</LangVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>


jmarolf:
@deinok what is the motivation for this change? Global tools should be forward compatible with newer .NET Core SDK

deinok:
In a `netcoreapp3.0` i can't add `dotnet-format` as a `<DotNetCliToolReference />`

jmarolf:
@deinok ah, I see.  In .NET Core 3 `DotNetCliToolReference` is being deprecated  for local tools (early docs [here](https://github.com/dotnet/cli/issues/10288)). in .NET Core 3 you will just need to specify installation of .NET format in a `dotnet-tools.json` file and `dotnet restore` will work automatically

src/dotnet-format.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFrameworks>netcoreapp2.1;netcoreapp2.2;netcoreapp3.0</TargetFrameworks>
```
jmarolf: we already specify that this tool can run on newer runtimes [here](https://github.com/dotnet/format/blob/master/src/runtimeconfig.template.json).  nuget not restoring it correcty in the 3.0 SDK is a bug that I am told is being fixed with the Local Tools SDK feature

wli3: > nuget not restoring it correcty in the 3.0 SDK is a bug

It should not be able to directly restored by nuget (as in adding to csproj and run restore). It is designed that way. You could use "tool restore" with local tools.

sharwell:
@deinok See https://github.com/dotnet/roslyn/pull/38562 for an example of this 😄

PR #288, opened 2019-07-27
Reuse compilation

jmarolf:


RikkiGibson:
Any ballpark on how much time this saves when there are no diagnostics to fix?

src/Analyzers/AnalyzerFormatter.cs
```diff
@@ -64,6 +64,11 @@ class AnalyzerFormatter : ICodeFormatter
                     {
                         logger.LogTrace($"Applying fixes for {codefix.GetType().Name}");
                         solution = await _applier.ApplyCodeFixesAsync(solution, result, codefix, logger, cancellationToken);
+                        var changedSolution = await _applier.ApplyCodeFixesAsync(solution, result, codefix, logger, cancellationToken);
+                        if (changedSolution.GetChanges(solution).Any())
```
sharwell: ❓ Why is the applier returning a different solution when no changes were made? I would only expect this condition to be false when `changedSolution` and `solution` are reference-equal.

JoeRobich: It would be great to simplify this and maybe we can.

Here is the code in the fix applier
https://github.com/dotnet/format/blob/35541706c9cfc94a769ff674b10bfdeddbd4c63c/src/Analyzers/SolutionCodeFixApplier.cs#L47-L50

PR #296, opened 2019-07-30
Do not run analyzers iteratively unless necessary

jmarolf:


jmarolf:
@RikkiGibson 	analyzer parallelism is something the analyzer needs to opt-into itself. all of roslyn analyzers should be opted in.

PR #305, opened 2019-08-02
no linq please

jmarolf:


sharwell:
:confused: why not?

JoeRobich:
There is still plenty of Linq in format. =) Jon was just iterating on his implementation of the analyzer runner.

PR #330, opened 2019-08-13
Add formatter for the charset .editorconfig option

JoeRobich:
`charset` support was requested by the community and should wrap up our support for the the core .editorconfig options.

### DIFF
--- a/src/Formatters/CharsetFormatter.cs
+++ b/src/Formatters/CharsetFormatter.cs
@@ -1,0 +1,71 @@
+﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information.
+
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Options;
+using Microsoft.CodeAnalysis.Text;
+using Microsoft.Extensions.Logging;
+using Microsoft.VisualStudio.CodingConventions;
+
+namespace Microsoft.CodeAnalysis.Tools.Formatters
+{
+    internal sealed class CharsetFormatter : DocumentFormatter
+    {
+        protected override string FormatWarningDescription => Resources.Fix_file_encoding;
+
+        private static Encoding Utf8 => new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+        private static Encoding Latin1 => Encoding.GetEncoding("iso-8859-1");
+
+        protected override Task<SourceText> FormatFileAsync(
+            Document document,
+            SourceText sourceText,
+            OptionSet options,
+            ICodingConventionsSnapshot codingConventions,
+            FormatOptions formatOptions,
+            ILogger logger,
+            CancellationToken cancellationToken)
+        {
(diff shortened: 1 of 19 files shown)


src/Formatters/CharsetFormatter.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public static Encoding GetCharset(string charsetOption)
+        {
+            switch (charsetOption)
+            {
+                case "latin1":
+                    return Latin1;
+                default:
+                case "utf-8":
```
RikkiGibson: it feels weird to have the default label mixed in with the case labels instead of being at the end.

Consider using a switch expression like the following

```cs
charsetOption switch
{
    "latin1" => Latin1,
    "utf-8-bom" => Encoding.UTF8,
    "utf-16be" => Encoding.BigEndianUnicode,
    "utf-16le" => Encoding.Unicode,
    _ => Utf8
}
```

tests/Formatters/AbstractFormatterTests.cs
```diff
@@ -66,50 +67,67 @@ protected AbstractFormatterTest()
...
+        /// <returns>The document contained within along with option set and coding conventions.</returns>
+        protected async Task<ImmutableArray<(Document, OptionSet, ICodingConventionsSnapshot)>> GetOnlyFileToFormatAsync(Solution solution, IReadOnlyDictionary<string, string> editorConfig)
+        {
+            var document = GetOnlyDocument(solution);
+            var options = (OptionSet)await document.GetOptionsAsync();
+
+            ICodingConventionsSnapshot codingConventions = new TestCodingConventionsSnapshot(editorConfig);
+            options = OptionsApplier.ApplyConventions(options, codingConventions, Language);
+
+            return new[] { (document, options, codingConventions) }.ToImmutableArray();
```
RikkiGibson: It looks like you were just migrating this existing code around, but please consider changing this to use `ImmutableArray.Create`

tests/Formatters/CharsetFormatterTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        [InlineData("utf-16le", "latin1")]
+        [InlineData("utf-16le", "utf-8")]
+        [InlineData("utf-16le", "utf-8-bom")]
+        [InlineData("utf-16le", "utf-16be")]
+        public async Task TestCharsetWrong_CharsetFixed(string codeValue, string expectedValue)
+        {
+            var codeEncoding = CharsetFormatter.GetCharset(codeValue);
+            var expectedEncoding = CharsetFormatter.GetCharset(expectedValue);
+
+            var testCode = "class C { }";
```
RikkiGibson: It might be worthwhile to verify that the encoding really changed the test code contents, perhaps by making the test class name a more complex character, and including a byte[] with the final raw content in the InlineData

JoeRobich: Yeah, we will need to make an integration test for that. There wouldn't be any change until the workspace has saved the document changes. Since these tests use AdhocWorkspace that isn't possible because it doesn't implement persistence.

PR #335, opened 2019-08-15
Add FolderWorkspace and flag to find files by folder

JoeRobich:
FolderWorkspace treats code files beneath the folder as part of the Solution. This workspace wouldn't be suitable to future analyzer/codefix work since semantic information wouldn't be available, but it does suffice to provide a faster path for whitespace formatting.

Opt-in to this experience by using the `--folder` option instead of `--workspace`.

- [x] Add FolderWorkspace
- [x] Add commandline option to use FolderWorkspace
- [ ] Update readme.md


JoeRobich:
@jmarolf please take a look

Issue #379, opened 2019-09-18
Dry run output format options


xt0rted:
I'm working on a tool that takes the output from a dry run and reports on it. Right now I'm having to parse this with a regex to get the file path, line number, and message. Is there a way to output this as json, or to write my own output formatter and use it as a plugin?

jmarolf:
seems reasonable to offer json output at the least.  I would accept a PR for this.

jpshrader:
Any additional thoughts on the structure of this? Just FileName/Path, line number, and message?

I can take a crack at this when I find some time.

jmarolf:
no objections from me. I would happily accept a pr here

jpshrader:
I've got a branch ready, but looks like I have a permissions issue

jmarolf:
thanks @jpshrader!You should be able to [fork ](https://help.github.com/en/github/getting-started-with-github/fork-a-repo)this repo and [create a pull request](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork) without any additional permissions. This is the workflow that I have used here. Is this not working for you?

jpshrader:
@jmarolf Ah, okay, gotcha. I was trying to push to a branch in this repo. Thanks! Apologies for the few days delay, I was only release duty last week.

I've opened the [PR here](https://github.com/dotnet/format/pull/495). I added some comments on some questions for you maintainers, so feel free to leave some red ink on it.

JoeRobich:
This functionality was implemented in #495

PR #495, opened 2020-01-21
format-379: Add `--report` command line argument to export json format report to given directory 

jpshrader:
PR for [this task](https://github.com/dotnet/format/issues/379). I diverged a tad from the original ask (generate a report when the `--dry-run` is provided) and decided that this would probably be useful in both normal and `--dry-run` modes. Also allowed the user to provide the output directory for the report, which I felt would be handy too (e.g. publishing it to a report viewer in Jenkins).

### DIFF
--- a/src/CodeFormatter.cs
+++ b/src/CodeFormatter.cs
@@ -69,7 +69,7 @@ namespace Microsoft.CodeAnalysis.Tools
 
                 var formattedFiles = new List<FormattedFile>();
                 var formattedSolution = await RunCodeFormattersAsync(
-                    solution, formatableFiles, options, logger, cancellationToken, formattedFiles).ConfigureAwait(false);
+                    solution, formatableFiles, options, logger, formattedFiles, cancellationToken).ConfigureAwait(false);
 
                 var formatterRanMS = workspaceStopwatch.ElapsedMilliseconds - loadWorkspaceMS - determineFilesMS;
                 logger.LogTrace(Resources.Complete_in_0_ms, formatterRanMS);
@@ -225,14 +225,14 @@ namespace Microsoft.CodeAnalysis.Tools
             ImmutableArray<(DocumentId, OptionSet, ICodingConventionsSnapshot)> formattableDocuments,
             FormatOptions options,
             ILogger logger,
-            CancellationToken cancellationToken,
-            List<FormattedFile> formattedFiles)
+            List<FormattedFile> formattedFiles,
+            CancellationToken cancellationToken)
         {
             var formattedSolution = solution;
 
             foreach (var codeFormatter in s_codeFormatters)
             {
-                formattedSolution = await codeFormatter.FormatAsync(formattedSolution, formattableDocuments, options, logger, cancellationToken, formattedFiles).ConfigureAwait(false);
+                formattedSolution = await codeFormatter.FormatAsync(formattedSolution, formattableDocuments, options, logger, formattedFiles, cancellationToken).ConfigureAwait(false);
             }
 
             return formattedSolution;
--- a/src/CodeFormatter.cs
+++ b/src/CodeFormatter.cs
@@ -6,17 +6,16 @@ using System.Collections.Immutable;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text.Json;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.MSBuild;
 using Microsoft.CodeAnalysis.Options;
-using Microsoft.CodeAnalysis.Tools.Utilities;
 using Microsoft.CodeAnalysis.Tools.Formatters;
+using Microsoft.CodeAnalysis.Tools.Utilities;
 using Microsoft.CodeAnalysis.Tools.Workspaces;
 using Microsoft.Extensions.Logging;
 using Microsoft.VisualStudio.CodingConventions;
-using System.Text;
-using Newtonsoft.Json;
 
 namespace Microsoft.CodeAnalysis.Tools
 {
@@ -68,13 +67,14 @@ namespace Microsoft.CodeAnalysis.Tools
 
                 logger.LogTrace(Resources.Running_formatters);
 
+                var formattedFiles = new List<FormattedFile>();
                 var formattedSolution = await RunCodeFormattersAsync(
-                    solution, formatableFiles, options, logger, cancellationToken).ConfigureAwait(false);
+                    solution, formatableFiles, options, logger, cancellationToken, formattedFiles).ConfigureAwait(false);
 
                 var formatterRanMS = workspaceStopwatch.ElapsedMilliseconds - loadWorkspaceMS - determineFilesMS;
                 logger.LogTrace(Resources.Complete_in_0_ms, formatterRanMS);
 
-                var solutionChanges = formattedSolution.Solution.GetChanges(solution);
+                var solutionChanges = formattedSolution.GetChanges(solution);
 
                 var filesFormatted = 0;
                 foreach (var projectChanges in solutionChanges.GetProjectChanges())
@@ -89,7 +89,7 @@ namespace Microsoft.CodeAnalysis.Tools
 
                 var exitCode = 0;
 
-                if (saveFormattedFiles && !workspace.TryApplyChanges(formattedSolution.Solution))
+                if (saveFormattedFiles && !workspace.TryApplyChanges(formattedSolution))
                 {
                     logger.LogError(Resources.Failed_to_save_formatting_changes);
                     exitCode = 1;
@@ -97,14 +97,16 @@ namespace Microsoft.CodeAnalysis.Tools
 
                 if (exitCode == 0 && !string.IsNullOrWhiteSpace(reportPath))
                 {
-                    var reportName = "format-report.json";
-                    var reportFilePath = new StringBuilder(reportPath);
-                    reportFilePath = reportPath.EndsWith('\\') ? reportFilePath.Append(reportName) : reportFilePath.Append($"\\{reportName}");
+                    var reportFilePath = GetReportFilePath(reportPath);
 
                     logger.LogInformation(Resources.Writing_formatting_report_to_0, reportFilePath);
-                    var formattedFilesJson = JsonConvert.SerializeObject(formattedSolution.FormattedFiles, Newtonsoft.Json.Formatting.Indented);
+                    var seralizerOptions = new JsonSerializerOptions
+                    {
+                        WriteIndented = true
+                    };
+                    var formattedFilesJson = JsonSerializer.Serialize(formattedFiles, seralizerOptions);
 
-                    File.WriteAllText(reportFilePath.ToString(), formattedFilesJson);
+                    File.WriteAllText(reportFilePath, formattedFilesJson);
                 }
 
                 logger.LogDebug(Resources.Formatted_0_of_1_files, filesFormatted, fileCount);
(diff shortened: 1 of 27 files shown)


src/dotnet-format.csproj
```diff
@@ -29,6 +29,7 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsDependencyInjectionVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(MicrosoftExtensionsLoggingVersion)" />
     <PackageReference Include="Microsoft.VisualStudio.CodingConventions" Version="$(MicrosoftVisualStudioCodingConventionsVersion)" Publish="true" />
+    <PackageReference Include="Newtonsoft.Json" Version="$(NewtonsoftJsonVerison)" />
```
jpshrader: I could have used [System.Text.Json](https://docs.microsoft.com/en-us/dotnet/api/system.text.json?view=netcore-3.0) for this, but that requires .Net Core 3.0.

I could upgrade the .Net Core version if you'd like, but didn't want to do that unilaterally.

tests/Formatters/AbstractFormatterTests.cs
```diff
@@ -91,16 +91,25 @@ private protected async Task<SourceText> TestAsync(string testCode, string expec
...
             var filesToFormat = await GetOnlyFileToFormatAsync(solution, editorConfig);
 
             var formattedSolution = await Formatter.FormatAsync(solution, filesToFormat, formatOptions, Logger, default);
-            var formattedDocument = GetOnlyDocument(formattedSolution);
+            var formattedDocument = GetOnlyDocument(formattedSolution.Solution);
             var formattedText = await formattedDocument.GetTextAsync();
 
             Assert.Equal(expectedCode, formattedText.ToString());
 
+            if (testCode != expectedCode)
```
jpshrader: `if`s in tests are a no-no, but this would require a fairly moderate refactor of this structure. 

1. Pass an additional boolean param to the `TestAsync` function, `expectingFileChanges`, which will replace this check.
2. Break this out into its own fixture, which tests the `FormattedFiles` behaviour in isolation.

Personally, I'd prefer #2, but I'll defer to you. Any particular preference?

src/Formatters/ICodeFormatter.cs
```diff
@@ -14,7 +15,7 @@ internal interface ICodeFormatter
         /// <summary>
         /// Applies formatting and returns a formatted <see cref="Solution"/>.
         /// </summary>
-        Task<Solution> FormatAsync(
+        Task<(Solution Solution, List<FormattedFile> FormattedFiles)> FormatAsync(
```
jpshrader: Another instance of style choice. I saw some precedent for using tuples, so decided to use it here. If you would prefer that I created a new class for this, let me know.

jpshrader:
Below is an example output of this change. Let me know if you'd prefer this be converted to camelCase.

```
[
  {
    "DocumentId": {
      "ProjectId": {
        "Id": "18cf4766-380a-4dc6-ab56-bc1e3b7625a8"
      },
      "Id": "81d1c87b-2187-4a7c-905d-9c413b17f0af"
    },
    "FileName": "CodeFormatter.cs",
    "FilePath": "C:\\projects\\format\\src\\CodeFormatter.cs",
    "FileChanges": [
      {
        "LineNumber": 291,
        "CharNumber": 30,
        "FormatDescription": "Fix whitespace formatting."
      }
    ]
  }
]
```

jmarolf:
@JoeRobich ptal

src/dotnet-format.csproj
jmarolf: I would appreciate it if you upgraded the project to 3.1 and useing System.Text.Json

src/CodeFormatter.cs
```diff
@@ -87,12 +89,24 @@ internal static class CodeFormatter
...
+                if (saveFormattedFiles && !workspace.TryApplyChanges(formattedSolution.Solution))
                 {
                     logger.LogError(Resources.Failed_to_save_formatting_changes);
                     exitCode = 1;
                 }
 
+                if (exitCode == 0 && !string.IsNullOrWhiteSpace(reportPath))
+                {
+                    var reportName = "format-report.json";
+                    var reportFilePath = new StringBuilder(reportPath);
```
jmarolf: I don't think we need a `StringBuilder`to concatinate two strings

src/CodeFormatter.cs
```diff
@@ -15,6 +15,8 @@
 using Microsoft.CodeAnalysis.Tools.Workspaces;
 using Microsoft.Extensions.Logging;
 using Microsoft.VisualStudio.CodingConventions;
+using System.Text;
```
jmarolf: sort usings

src/CodeFormatter.cs
```diff
@@ -187,18 +201,20 @@ private static void LogWorkspaceDiagnostics(ILogger logger, bool logWorkspaceWar
...
         {
-            var formattedSolution = solution;
+            var formattedSolution = (Solution: solution, FormattedFiles: new List<FormattedFile>());
 
             foreach (var codeFormatter in s_codeFormatters)
             {
-                formattedSolution = await codeFormatter.FormatAsync(formattedSolution, formattableDocuments, options, logger, cancellationToken).ConfigureAwait(false);
+                var formatterResponse = await codeFormatter.FormatAsync(formattedSolution.Solution, formattableDocuments, options, logger, cancellationToken).ConfigureAwait(false);
+                formattedSolution.Solution = formatterResponse.Solution;
+                formattedSolution.FormattedFiles.AddRange(formatterResponse.FormattedFiles);
```
jmarolf: Why not just give the code formatter a list and have it add its formatted files to it instead of returning them in a tuple?

tests/Formatters/AbstractFormatterTests.cs
jmarolf: I would also prefer number 2, thanks!

jmarolf:
@jpshrader looks like a fantastic start! Thanks for the controbution! I left a few comments and I want @JoeRobich to take a look before merging

src/FileChange.cs
```diff
@@ -0,0 +1,21 @@
+﻿using Microsoft.CodeAnalysis.Text;
+
+namespace Microsoft.CodeAnalysis.Tools
+{
+    public class FileChange
+    {
+        public int LineNumber { get; set; }
```
JoeRobich: Can we remove the setters if they aren't necessary?

src/FormattedFile.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Collections.Generic;
+
+namespace Microsoft.CodeAnalysis.Tools
+{
+    public class FormattedFile
+    {
+        public DocumentId DocumentId { get; set; }
```
JoeRobich: Can we remove these setters if they aren't necessary?

src/Program.cs
```diff
@@ -36,13 +36,14 @@ private static async Task<int> Main(string[] args)
                 .AddOption(new Option(new[] { "--dry-run" }, Resources.Format_files_but_do_not_save_changes_to_disk, new Argument<bool>()))
                 .AddOption(new Option(new[] { "--check" }, Resources.Terminate_with_a_non_zero_exit_code_if_any_files_were_formatted, new Argument<bool>()))
                 .AddOption(new Option(new[] { "--files" }, Resources.A_comma_separated_list_of_relative_file_paths_to_format_All_files_are_formatted_if_empty, new Argument<string>(() => null)))
+                .AddOption(new Option(new[] { "--report" }, Resources.Accepts_a_file_path_which_if_provided_will_produce_a_format_report_json_file_in_the_given_directory, new Argument<string>(() => null)))
```
JoeRobich: If we made the default value ".", then if just the option is provided it would drop the report in the current folder.

src/CodeFormatter.cs
```diff
@@ -87,12 +89,24 @@ internal static class CodeFormatter
...
                 var exitCode = 0;
 
-                if (saveFormattedFiles && !workspace.TryApplyChanges(formattedSolution))
+                if (saveFormattedFiles && !workspace.TryApplyChanges(formattedSolution.Solution))
                 {
                     logger.LogError(Resources.Failed_to_save_formatting_changes);
                     exitCode = 1;
                 }
 
+                if (exitCode == 0 && !string.IsNullOrWhiteSpace(reportPath))
```
JoeRobich: If `reportPath` is a path to a file (".\\logs\\formatting.json"), can we use that as the report path? otherwise, appending a default file name to path is desirable.

JoeRobich:
@jpshrader Thanks for working on this. It is a good start to providing more report options such as patch file support.

src/dotnet-format.csproj
jmarolf: this is actually more involved that I though. I've done it in this PR: https://github.com/dotnet/format/pull/496

jmarolf: actually scratch that. You should be able to reference the nuget package for `System.Text.Json` and use it already

src/Program.cs
jpshrader: I think there may be difficulty distinguishing between providing an empty argument vs not providing the option at all (unless I'm missing something). Changing the default argument to "." will set 'reportPath` even if you don't provide the argument at all.

jmarolf:
@jpshrader looks good to me! If @JoeRobich signs off I'll merge it

tests/Formatters/FormattedFilesTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            var document = project.Documents.Single();
+            var formatOptions = new FormatOptions(
+                workspaceFilePath: project.FilePath,
+                workspaceType: WorkspaceType.Folder,
+                logLevel: LogLevel.Trace,
+                saveFormattedFiles: false,
+                changesAreErrors: false,
+                filesToFormat: ImmutableHashSet.Create(document.FilePath),
+                reportPath: string.Empty);
+            var editorConfig = new Dictionary<string, string>()
```
JoeRobich: Extract this to a readonly field since it will apply to all tests.

src/Resources.resx
```diff
@@ -210,4 +210,10 @@
   <data name="Cannot_specify_both_folder_and_workspace_options" xml:space="preserve">
     <value>Cannot specify both folder and workspace options.</value>
   </data>
+  <data name="Accepts_a_file_path_which_if_provided_will_produce_a_format_report_json_file_in_the_given_directory" xml:space="preserve">
+    <value>Accepts a file path, which if provided, will produce a `format-report.json` file in the given directory.</value>
```
JoeRobich: Since the filename won't always be `format-report.json` maybe change this to be `json report`

src/Formatters/DocumentFormatter.cs
```diff
@@ -26,10 +27,11 @@ internal abstract class DocumentFormatter : ICodeFormatter
             ImmutableArray<(DocumentId, OptionSet, ICodingConventionsSnapshot)> formattableDocuments,
             FormatOptions formatOptions,
             ILogger logger,
-            CancellationToken cancellationToken)
+            CancellationToken cancellationToken,
+            List<FormattedFile> formattedFiles)
```
JoeRobich: Please leave `cancellationToken` as the last parameter everywhere it was moved.

jmarolf:
@JoeRobich looks good to merge to me!

jpshrader:
Awesome, thanks @JoeRobich / @jmarolf ! Appreciate the reviews!

PR #496, opened 2020-01-21
Move us to .NET Core 3.1

jmarolf:


src/dotnet-format.csproj
```diff
@@ -1,9 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
sharwell: ❔ Aside from limiting the set of users that can use this project, what's the advantage from this?

src/dotnet-format.csproj
```diff
@@ -37,6 +37,15 @@
...
     <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Workspaces" Version="$(MicrosoftCodeAnalysisVersion)" />
+
+    <PackageReference Include="Microsoft.Win32.Primitives" Version="$(MicrosoftWin32PrimitivesVersion)" />
+    <PackageReference Include="System.Diagnostics.Tracing" Version="$(SystemDiagnosticsTracingVersion)" />
+    <PackageReference Include="System.IO.FileSystem.Primitives" Version="$(SystemIOFileSystemPrimitivesVersion)" />
+    <PackageReference Include="System.IO.FileSystem" Version="$(SystemIOFileSystemVersion)" />
+    <PackageReference Include="System.Net.Primitives" Version="$(SystemNetPrimitivesVersion)" />
+    <PackageReference Include="System.Runtime.Handles" Version="$(SystemRuntimeHandlesVersion)" />
+    <PackageReference Include="System.Runtime.InteropServices" Version="$(SystemRuntimeInteropServicesVersion)" />
+    <PackageReference Include="System.Text.Encoding.Extensions" Version="$(SystemTextEncodingExtensionsVersion)" />
```
sharwell: 📝 All of these should be picked up automatically. We shouldn't need to manually reference them.

jmarolf: Coding conventions is net standard 1.6 which will cause a package downgrade otherwise

src/dotnet-format.csproj
jmarolf: System. Text. Json

sharwell: System.Text.Json is netstandard2.0. It should work without changes right?

global.json
```diff
@@ -1,6 +1,10 @@
 {
   "tools": {
-    "dotnet": "3.0.101"
+    "dotnet": "3.1.100"
+  },
+  "sdk": {
```
JoeRobich: Do we have to pin to a particular SDK?

jmarolf: You need at least `3.1` to build `3.1`. adding `"rollForward": "major"` should enable any newer sdk to be used without issue.

src/dotnet-format.csproj
jmarolf: well so it is. I feel silly now. I don't think we should do this then until 2.1 goes out of support

jmarolf:
@sharwell correctly points out that we should be able to get all the advantages of better libraries in `3.1` by just referencing their nuget packages. I don't think we want to make this move until 2.1 goes out of support

Issue #500, opened 2020-01-24
Support Folder and File Exclusions


JoeRobich:
Ability to list folder or files that should be excluded from formatting.

jpshrader:
@JoeRobich Mind if I take a crack at this?

JoeRobich:
@jpshrader That would be great. Start by posting your proposed design here and we can talk through it.

jpshrader:
@JoeRobich I was thinking it might look something like:

1. Accept a comma-separated list of file/folder paths (`--exclude <stuff>`), which like the `--report` option, is passed along to [CodeFormatter.cs](https://github.com/dotnet/format/blob/master/src/CodeFormatter.cs) via the `FormatOptions` object.

2. Add a similar `if` block to the [GetDocumentAndOptions](https://github.com/dotnet/format/blob/master/src/CodeFormatter.cs#L286-L289) method in `CodeFormatter`, which excludes files if they fall under any of the excluded files/folders.

That logic could also go a layer up in in the [DetermineFormattableFiles](https://github.com/dotnet/format/blob/master/src/CodeFormatter.cs#L249-L270), but I think the above approach is cleaner.

sandyarmstrong:
Personally, I would find it very useful to be able to specify exclusions in some sort of configuration file, so that using the tool locally or via CI would yield the same results without having to make a tedious list of arguments.

I do think an `--exclude <stuff>` approach is great. But I hope it's a step on the road to configuration-based formatting.

BTW how evil would it be to put this kind of configuration data in `global.json`? :)

JoeRobich:
> I would find it very useful to be able to specify exclusions in some sort of configuration file

@sandyarmstrong  I opened https://github.com/dotnet/format/issues/520 to track the configuration file proposal. Feel free to make suggestions.

jpshrader:
Reading through comments in #520, to @sandyarmstrong's [point](https://github.com/dotnet/format/issues/520#issuecomment-582669311):
>... to capture files that aren't part of the main solution, but then need to be able to avoid files (for example, template files, or certain other shared/copied files) that are not really "ours" to format.

It might be valuable to allow consumers to use wildcards in their paths to account for these types of situations (`*.template.cs`, for instance). I don't think the other options for folder/files support this, so apologies if this idea has been swatted down in the past.

JoeRobich:
@jpshrader The proposal looks good. I feel like we'll have to add an `include` alias for the `files` option. =)
[...]

PR #529, opened 2020-02-11
format-500: Add `--exclude` option to ignore given files/folders

jpshrader:
PR for: https://github.com/dotnet/format/issues/500


src/CodeFormatter.cs
```diff
@@ -347,5 +337,36 @@ private static void LogWorkspaceDiagnostics(ILogger logger, bool logWorkspaceWar
...
+            else if (!document.SupportsSyntaxTree)
+            {
+                return true;
+            }
+            else if (await GeneratedCodeUtilities.IsGeneratedCodeAsync(document, cancellationToken).ConfigureAwait(false))
+            {
+                // Ignore generated code files.
+                return true;
+            }
+            else if (!filesToIgnore.IsEmpty && filesToIgnore.Any(f => document.FilePath.Contains(f, StringComparison.OrdinalIgnoreCase)))
```
jpshrader: Do we want to have ignore casing on this string compare? If so, should I also add it on :347

src/Resources.resx
```diff
@@ -216,4 +216,7 @@
   <data name="Writing_formatting_report_to_0" xml:space="preserve">
     <value>Writing formatting report to: '{0}'.</value>
   </data>
+  <data name="A_comma_separated_list_of_relative_file_or_folder_paths_to_exclude_from_formatting" xml:space="preserve">
+    <value>A comma separated list of relative file/folder paths to exclude from formatting</value>
```
JoeRobich: If this ever gets localized, then it is good to avoid idioms that may not translate such as '/' to mean 'or'. 

```suggestion
    <value>A comma separated list of relative file or folder paths to exclude from formatting.</value>
```

src/CodeFormatter.cs
JoeRobich: Also, since the paths should be normalized, I think `.StartsWith` might be "more correct" here.

jpshrader: Changed it to `StartsWith`, but I still need to keep the `StringComparison` bit since we're doing `document.FilePath.StartsWith`. Unlike the `ImmutableHashset`, `document.FilePath`'s comparer doesn't ignore casing.

JoeRobich:
🎉

jpshrader:
Thanks @JoeRobich! If you've got anything you'd like a hand with, feel free to give me a shout.

PR #533, opened 2020-02-12
Enhance --files option

JoeRobich:
- Allow folder paths to be specified for inclusion via the `--files` option
- Add `--include` alias for `--files` option

### DIFF
--- a/src/Workspaces/FolderWorkspace_ProjectLoader.cs
+++ b/src/Workspaces/FolderWorkspace_ProjectLoader.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Immutable;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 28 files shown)


src/Workspaces/FolderWorkspace_ProjectLoader.cs
```diff
@@ -36,19 +37,15 @@ public virtual async Task<ProjectInfo> LoadProjectInfoAsync(string folderPath, I
...
+                    if (!pathsToInclude.IsEmpty)
                     {
-                        filePaths = Directory.GetFiles(folderPath, $"*{fileExtension}", SearchOption.AllDirectories);
-                    }
-                    else
-                    {
-                        filePaths = filesToInclude.Where(
-                            filePath => filePath.EndsWith(fileExtension) && File.Exists(filePath)).ToArray();
+                        filePaths = filePaths.Where(filePath =>
+                            pathsToInclude.Any(path => filePath.StartsWith(path, StringComparison.OrdinalIgnoreCase))).ToArray();
```
jmarolf: Is there a .NET api to compare file paths that works if the file system is case-sensitive (like most linux file systems are?). Not a blocker jsut curious

JoeRobich: Not that I can find. Nothing off Path at least. Could roll our own getting FileInfo and using its FullPath which I would hope to be cased properly.

PR #541, opened 2020-02-13
Combine --check and --dry-run into a single option.

JoeRobich:
The use case for `--check` (CI Validation) also uses `--dry-run`. Use cases for `--dry-run` could ignore exit code if formatting changes are found.

Considers `--check` as the canonical flag and `--dry-run` as an alias.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,8 @@ Options:
                      the current directory for one.
   --verbosity, -v    Set the verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and
                      diag[nostic]
-  --dry-run          Format files, but do not save changes to disk.
-  --check            Terminates with a non-zero exit code if any files were formatted.
+  --check, --dry-run Formats files without saving changes to disk. Terminates with a non-zero exit code if any files
+                     were formatted.
   --include, --files A comma separated list of relative file or folder paths to include in formatting. All files are
                      formatted if empty.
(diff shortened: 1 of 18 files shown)


jmarolf:
I approve of this change. My only question is "should we bump the version number as this is a breaking change?"

JoeRobich:
@jmarolf I plan to bump version numbers for each release. That being said. I should start doing it after a release...

JoeRobich:
Version bump in this PR https://github.com/dotnet/format/pull/543

jmarolf:
Shouldn't this be a major version bump though?

JoeRobich:
@jmarolf sure

xt0rted:
I have a GitHub Action that wraps this tool ([xt0rted/dotnet-format](https://github.com/xt0rted/dotnet-format)) and I'm about to release it but was wondering how this change will effect my usage of it first.

When linting I use `--check --dry-run` so I can let the user decide if the workflow should exit with an error or not, and when applying fixes I'm also using `--check` to know if the rest of the workflow should continue (committing & pushing changes back).

Now that `--check` and `--dry-run` have been merged is there a way for the caller to know if files were formatted? I guess I could add another step that uses git for this, but it'd be easier if the tool told me instead.

JoeRobich:
Hi @xt0rted, Our understanding was that --check and --dry-run were used together basically 100% of the time for the purpose of linting (once users realized they needed to have both). I would recommend checking git for unstaged changes if you are relying on the old behavior of --check. Sorry again for the change.

Issue #9, opened 2019-02-20
dotnet-format: accept a list of files to format (instead of a full workspace)


BhaaLseN:
**Version Used**: dotnet-format 3.0.0-beta4-19105-10 (70a2f514eb3932648f771b086834cb0a376fee20)

**Steps to Reproduce**:
1. Change a number of files (so they appear in VCS output, such as `git diff --name-only`)
2. Pass that file list as command line parameters into `dotnet format` (or try to pipe them into stdin)

**Expected Behavior**:
`dotnet format` only changes the passed files instead of the whole project (or solution)

**Actual Behavior**:
`dotnet format` errors out with either "_Both a MSBuild project file and solution file found in $path. Specify which one to use with the --workspace option_" or "_Unrecognized command or argument 'path/to/file.cs'_". Or somewhat...not sure if I should call it _worse_ or just _different_: it runs successfully, formatting all files referenced by the project or solution (in case only a single project or solution file is found in the current working directory).

**Intended Use-Case**:
We run a few things as part of our CI (such as static analysis), and we'd also like to use `dotnet format` as a way to enforce code style before it even reaches the first review by a human. Right now this isn't too feasible as `dotnet format` takes quite a while on larger solutions with a lot of files; and running it only on the changed files would significantly improve the time to feedback (with the intention to simply error out in case it produces any changes in the CI working copy).

For smaller change sets, passing file names as command line parameter is likely a usable solution, but the larger the change set the more likely it would be to reach a length limit for arguments; so piping it into stdin would probably be the better approach there.

Or, alternatively (if it makes sense to do so), support this use case from within `dotnet format` itself (or a tool derived from it), along with the `--dry-run` switch: stop processing as soon as a change would have been made and return an error code that can be checked by a caller.

Thoughts and ideas are more than welcome, especially in case I missed some other tooling that already does this.

BhaaLseN:
Oh, nice, didn't know there was a repository dedicated to `dotnet format`; created the issue on `roslyn` since it had other related issues.
Thanks for moving it 👍

jmarolf:
@BhaaLseN 	you did the right thing. I just put this repo together recently

challen0:
@jmarolf @JoeRobich I was thinking of looking into this, as it's something I could use too. Is the way it's written in the above intended use case the way to move forward? Is a `--files` option necessary or would providing paths as arguments to the command be enough?

BhaaLseN:
[...]

PR #551, opened 2020-02-16
Use space-separated paths instead of comma-separated

am11:
#### Problem

1. in Windows as well as in Unix-like operating systems, if the file path contains comma, we cannot reliably use `--files` option.
2. we are currently incompatible with shell globbing expansion; without using additional tools as workaround (such as https://github.com/dotnet/format/issues/9#issuecomment-541097833).

#### Solution

Change the semantics of `--include` or `--files` and `--exclude` options from `comma-separated` to `space-separated`.

Now we can do things like `dotnet-format --files {src,test}/**/*.cs` or
`git ls-files :/{src,tests}/*.cs | xargs dotnet-format --files`.

#### Additional change

Change the output message to include full path to file, so we can distinguish the clashing files.

```diff
- dotnet-format --files ./fileA.cs,"a sub directory/fileA.cs"
+ dotnet-format --files ./fileA.cs "a sub directory/fileA.cs"

-  Formatted code file 'fileA.cs'.
-  Formatted code file 'fileA.cs'.
+  Formatted code file '/path/fileA.cs'.
+  Formatted code file '/path/a sub directory/fileA.cs'.
```

### DIFF
--- a/src/Program.cs
+++ b/src/Program.cs
@@ -34,8 +34,8 @@ namespace Microsoft.CodeAnalysis.Tools
                 .UseExceptionHandler()
                 .AddOption(new Option(new[] { "--folder", "-f" }, Resources.The_folder_to_operate_on_Cannot_be_used_with_the_workspace_option, new Argument<string>(() => null)))
                 .AddOption(new Option(new[] { "--workspace", "-w" }, Resources.The_solution_or_project_file_to_operate_on_If_a_file_is_not_specified_the_command_will_search_the_current_directory_for_one, new Argument<string>(() => null)))
-                .AddOption(new Option(new[] { "--include", "--files" }, Resources.A_comma_separated_list_of_relative_file_or_folder_paths_to_include_in_formatting_All_files_are_formatted_if_empty, new Argument<string>(() => null)))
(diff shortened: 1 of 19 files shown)


JoeRobich:
I like this change. I wish it had worked this way from the start. Could you update the  readme docs that use these options to match the new syntax as part of this PR.

am11:
@JoeRobich, updated the readme to reflect changes. :)

src/Program.cs
```diff
@@ -33,11 +33,11 @@ private static async Task<int> Main(string[] args)
                 {
                     Argument = new Argument<string>(() => null)
                 },
-                new Option(new[] { "--include", "--files" }, Resources.A_comma_separated_list_of_relative_file_or_folder_paths_to_include_in_formatting_All_files_are_formatted_if_empty)
+                new Option(new[] { "--include", "--files" }, Resources.A_list_of_relative_file_or_folder_paths_to_include_in_formatting_All_files_are_formatted_if_empty)
                 {
                     Argument = new Argument<string>(() => null)
```
am11: @JoeRobich, this was changed to `string[]` and also the `--exclude` case below.

JoeRobich: That is what I get for trying to fix merge conflicts using the GH editor. sorry

am11: No problem, thanks for the merge. I think API should error out in this case. Opened https://github.com/dotnet/command-line-api/issues/787 for discussion.

Issue #520, opened 2020-02-05
Support configuring dotnet-format with a configuration file


JoeRobich:
As we add more options to the configuration it is increasingly tedious to type them all out at the command line. 

- We should add support for reading in options from a "format.json" file. 
- We should look for the configuration file in the current working directory.
- We should support multiple configurations stored as top-level properties. The configuration can be specified from the command line with a new `-c|--configuration {name}` option. A "default" option can be configured which will be applied whenever a configuration is not specified.
- The option's property name should match the long name for the option.
- Options that support multiple values provided as a comma separated string can be expressed as either a comma separated string or as an array of values.
- Options are optional in the configuration and will take their default value if missing or if `null` is specified for their value.

Example "format.json":
```json
{
    "default": {
        "folder": "./src",
        "workspace": null,
        "dry-run": false,
        "check": false,
    },
    "all": {
        "workspace": "format.sln",
        "dry-run": false,
        "check": false,
        "files": null
    },
    "validate": {
        "folder": "./src",
        "workspace": null,
        "dry-run": true,
        "check": true,
        "files": [ ]
    }
}
```

sandyarmstrong:
My only addition here would be to make it a hidden file similar to `.gitconfig` and `.editorconfig`. I'd suggest `.dotnetformatconfig` or `.formatconfig`.

I really like the multiple configurations idea here. Cool stuff!

jmarolf:
I am not opposed to this in theory. My main concern is that we create an alternative to `.editorconfig`. In most cases I would expect someone to update the `.editorconfig` in their project to exclude/include files. I can see the rare case where I have some set of submodules that I need to configure, but my hope is that we first find a way to specify formatting options so that `dotnet format`, Visual Studio for Mac, VS Code, and Visual Studio can all do the same thing.

All those concerns aside, I would tentatively accept a PR for this.

sandyarmstrong:
My desires here are motivated by VSmac's build needs:

* We have several submodules, so there's that. We can't trivially use this tool as a CI check as long as we have submodules without perfect formatting. (the new `--report` option would at least allow us to process the output and do our own path exclusions...so far the output is only file names so we can't filter based on that)
* We would like to use the `--folder` option to capture files that aren't part of the main solution, but then need to be able to avoid files (for example, template files, or certain other shared/copied files) that are not really "ours" to format.

jmarolf:
[...]

PR #552, opened 2020-02-16
Update versions of dependencies

JoeRobich:
The latest version of System.CommandLine supports loading options via [response files](https://github.com/dotnet/command-line-api/wiki/Features-overview#response-files). This could be our solution to the request to support a configuration file (https://github.com/dotnet/format/issues/520).

### DIFF
--- a/src/Logging/NullScope.cs
+++ b/src/Logging/NullScope.cs
@@ -1,0 +1,19 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.CodeAnalysis.Tools.Logging
+{
+    internal class NullScope : IDisposable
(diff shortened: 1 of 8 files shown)


JoeRobich:
@jmarolf PTAL

src/Logging/NullScope.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.CodeAnalysis.Tools.Logging
+{
+    internal class NullScope : IDisposable
```
jmarolf: what is this used for?

JoeRobich: Used by the Loggers. It used to be a public type but has disappeared from the new version of the library.

PR #555, opened 2020-02-16
Support file globbing in --include and --exclude options

JoeRobich:
Use the Microsoft.Extensions.FileSystemGlobbing package to handle matching against the included and excluded paths.


JoeRobich:
@jmarolf PTAL

jmarolf:
As long as `Perf.cmd` works on your machine I have no concerns

JoeRobich:
Seemed to run fine. Thanks!

PR #590, opened 2020-03-04
Move to Roslyn's editorconfig support

JoeRobich:


JoeRobich:
@jmarolf please take a look

src/Utilities/EditorConfigOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            if (_backing.ContainsKey(key))
+            {
+                value = _backing[key];
+                return true;
+            }
+
+            value = default;
+            return false;
+        }
```
sharwell:
```suggestion
            => _backing.TryGetValue(key, out value);
```

src/Workspaces/FolderWorkspace_ProjectLoader.cs
```diff
@@ -27,14 +28,24 @@ public virtual async Task<ProjectInfo> LoadProjectInfoAsync(string folderPath, M
                     return null;
                 }
 
+                var editorConfigDocuments = EditorConfigFinder.GetEditorConfigPaths(folderPath)
+                    .Select(path =>
+                        DocumentInfo.Create(
+                            DocumentId.CreateNewId(projectId),
+                            name: ".editorconfig",
+                            loader: new FileTextLoader(path, DefaultEncoding),
```
sharwell: **.editorconfig** is UTF-8 by definition.

```suggestion
                            loader: new FileTextLoader(path, Encoding.UTF8),
```

src/Utilities/EditorConfigFinder.cs
```diff
@@ -0,0 +1,40 @@
...
+            while (directory.Parent is object)
+            {
+                directory = directory.Parent;
+
+                editorConfigPaths.AddRange(
+                    directory.GetFiles(".editorconfig", SearchOption.TopDirectoryOnly)
+                        .Select(file => file.FullName));
+            }
+
+            return editorConfigPaths.OrderBy(filePath => filePath.Length).ToImmutableArray();
```
sharwell: > OrderBy(filePath => filePath.Length)

Does the compiler do this automatically?

tests/Formatters/AbstractFormatterTests.cs
```diff
@@ -120,15 +119,13 @@ private protected async Task<SourceText> TestAsync(string testCode, string expec
         /// <param name="solution">A Solution containing a single Project containing a single Document.</param>
         /// <param name="editorConfig">The editorconfig to apply to the documents options set.</param>
         /// <returns>The document contained within along with option set and coding conventions.</returns>
-        protected async Task<ImmutableArray<(DocumentId, OptionSet, ICodingConventionsSnapshot)>> GetOnlyFileToFormatAsync(Solution solution, IReadOnlyDictionary<string, string> editorConfig)
+        protected async Task<ImmutableArray<(DocumentId, OptionSet, AnalyzerConfigOptions)>> GetOnlyFileToFormatAsync(Solution solution, IReadOnlyDictionary<string, string> editorConfig)
         {
             var document = GetOnlyDocument(solution);
             var options = (OptionSet)await document.GetOptionsAsync();
+            var analyzerConfigOptions = (AnalyzerConfigOptions)new EditorConfigOptions(editorConfig);
```
sharwell: 💡 Prefer implicit casts to explicit casts

```suggestion
            AnalyzerConfigOptions analyzerConfigOptions = new EditorConfigOptions(editorConfig);
```

src/CodeFormatter.cs
```diff
@@ -30,6 +30,8 @@ internal static class CodeFormatter
             new CharsetFormatter(),
         }.ToImmutableArray();
 
+        private static readonly Dictionary<ProjectId, AnalyzerConfigSet> s_analyzerConfigSets = new Dictionary<ProjectId, AnalyzerConfigSet>();
```
sharwell: 💡 Consider using `ConditionalWeakTable<Project, AnalyzerConfigSet>` to avoid unbounded memory growth if used in a long-lived process.

src/Utilities/EditorConfigFinder.cs
JoeRobich: Indeed. Will remove

JoeRobich:
@sharwell @jmarolf This is ready for another review.

src/CodeFormatter.cs
jmarolf: skeptical that this will matter for a commandline tool

---
In reply to: [393801704](https://github.com/dotnet/format/pull/590#discussion_r393801704) [](ancestors = 393801704)

src/CodeFormatter.cs
```diff
@@ -293,99 +289,105 @@ private static void LogWorkspaceDiagnostics(ILogger logger, bool logWorkspaceWar
...
-            else if (!includeGeneratedFiles && await GeneratedCodeUtilities.IsGeneratedCodeAsync(document, cancellationToken).ConfigureAwait(false))
-            {
-                // Ignore generated code files.
-                return true;
-            }
-            else
+
+            var parseConfigs = project.AnalyzerConfigDocuments.Select(ParseAnalyzerConfigAsync);
+            var configs = await Task.WhenAll(parseConfigs);
+            configSet = AnalyzerConfigSet.Create(configs);
```
jasonmalinowski, line 381: Do we need to expose a public API to get the workspace-produced AnalyzerConfigSet? If that's internal we should just make it public, or make a method on Document that returns the document-specific AnalyzerConfig.

src/CodeFormatter.cs
```diff
@@ -293,99 +289,105 @@ private static void LogWorkspaceDiagnostics(ILogger logger, bool logWorkspaceWar
...
+                    }
                 }
-
-                addedFilePaths.Add(document.FilePath);
-                formattableFiles.Add((document.Id, options, codingConventions));
             }
 
-            return (fileCount, formattableFiles.ToImmutableArray());
+            // If no files are covered by an editorconfig, then return them all. Otherwise only return
+            // files that are covered by an editorconfig.
```
jasonmalinowski, line 341: I admit I'm not understanding why here -- can the comment clarify?

src/CodeFormatter.cs
```diff
@@ -293,99 +289,105 @@ private static void LogWorkspaceDiagnostics(ILogger logger, bool logWorkspaceWar
...
+                    if (!includeGeneratedFiles &&
+                        await GeneratedCodeUtilities.IsGeneratedCodeAsync(syntaxTree, cancellationToken).ConfigureAwait(false))
+                    {
+                        continue;
+                    }
+
+                    var analyzerConfigOptions = await GetAnalyzerConfigOptionsAsync(document, cancellationToken).ConfigureAwait(false);
+                    var optionSet = await document.GetOptionsAsync(cancellationToken).ConfigureAwait(false);
+
+                    var formattableDocument = new DocumentWithOptions(document, optionSet, analyzerConfigOptions);
```
jasonmalinowski, line 326: Why is DocumentWithOptions needed if we can fetch the OptionSet when you need it later or potentially the AnalyzerConfigOptions too? Is this just a perf optimization or needed for something else?

eng/Versions.props
```diff
@@ -24,10 +24,12 @@
     <MicrosoftExtensionsDependencyInjectionVersion>$(MicrosoftExtensionsVersion)</MicrosoftExtensionsDependencyInjectionVersion>
     <MicrosoftExtensionsFileSystemGlobbingVersion>$(MicrosoftExtensionsVersion)</MicrosoftExtensionsFileSystemGlobbingVersion>
     <MicrosoftExtensionsLoggingVersion>$(MicrosoftExtensionsVersion)</MicrosoftExtensionsLoggingVersion>
-    <MicrosoftVisualStudioCodingConventionsVersion>1.1.20180503.2</MicrosoftVisualStudioCodingConventionsVersion>
     <MicrosoftCodeAnalysisAnalyzersVersion>3.0.0</MicrosoftCodeAnalysisAnalyzersVersion>
     <MicrosoftCodeAnalysisVersion>$(MicrosoftNETCoreCompilersPackageVersion)</MicrosoftCodeAnalysisVersion>
     <SystemTextJsonVersion>4.7.0</SystemTextJsonVersion>
+
+    <!-- Resolves the inability to load the NugetSdkResolver from the .NET 5 SDK -->
```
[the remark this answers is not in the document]
jmarolf: I don't know how this can ever be resolved since this has to do with how MSBuild loads dependencies for targets. Though @rainersigwald can correct me if I am wrong.

src/CodeFormatter.cs
JoeRobich: I will add one. Initially we would format all document, even if the file wasn't covered by and .editorconfig and would have defaults applied. This behavior was an early requested change since users were surprised to have files not specified by the .editorconfig modified. My assumption is that users without an .editorconfig still wanted formatting (they did run a formatter after all), so it runs on all files with defaults.

src/CodeFormatter.cs
JoeRobich: I thought I was saving work, but you are right. Since nothing in the solution is changing and these are cached within the DocumentState and AnalyzerConfigSet respectively. This could just return document ids and we could fetch everything again later.

eng/Versions.props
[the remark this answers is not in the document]
JoeRobich: I'll add a link to the bug

src/CodeFormatter.cs
JoeRobich: I'll take this suggestion as a follow up refactoring. https://github.com/dotnet/format/issues/691

PR #645, opened 2020-04-14
Make the --folder options the default workspace option.

### LINKED ISSUE #644: Have the --folder option be the default
Today there are a few problems with using MSBuildWorkspace

1. It is slow
2. It has very complex dll loading logic
3. It can cause spurious errors and basically requires that a project be fully build-able via an MSBuildWorkspace

The folder workspace by comparison has none of these problems. I propose that we bump the major version and have the tool default to the `--folder` option.

Future features, such as running analyzers will necessitate the use of MSBuildWorkspace workspace but I expect us to fall back to using MSBuildWorkspace only when absolutely necessary.

JoeRobich:
Change the behavior from looking for a .csproj or .sln in the current directory when no folder or workspace option is provided to using the current directory as a folder workspace.

Resolves #644

### DIFF
--- a/src/Program.cs
+++ b/src/Program.cs
@@ -96,15 +96,7 @@ namespace Microsoft.CodeAnalysis.Tools
                     logger.LogWarning(Resources.Cannot_specify_both_folder_and_workspace_options);
                     return 1;
                 }
-
-                if (!string.IsNullOrEmpty(folder))
-                {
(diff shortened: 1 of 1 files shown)


JoeRobich:
The version was bumped earlier and we haven't had a release since.

jmarolf:
ah, no worries then

PR #672, opened 2020-05-12
Move to .NET Core 3.1

JoeRobich:
I revived this PR (https://github.com/dotnet/format/pull/496) to fix an issue we've been experiencing with .NET 5 SDK previews.

The Issue:
> Msbuild failed when processing the file 'C:\Users\jorobich\source\repos\roslyn\src\Deployment\RoslynDeployment.csproj' with message: C:\Program Files\dotnet\sdk\5.0.100-preview.2.20169.1\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.DefaultItems.targets: (257, 5): The "CheckForImplicitPackageReferenceOverrides" task could not be loaded from the assembly C:\Program Files\dotnet\sdk\5.0.100-preview.2.20169.1\Sdks\Microsoft.NET.Sdk\targets\..\tools\netcoreapp3.1/Microsoft.NET.Build.Tasks.dll. Could not load file or assembly 'System.Runtime, Version=4.2.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.

The Cause:
When running against a .NET 5 SDK preview, the Build.Tasks require System.Runtime 4.2.2.0. Since dotnet-format is built against netcoreapp2.1 and has roll forward enabled, If there is a 2.x or 3.0 runtime installed, then at runtime there is already a lower versioned version of System.Runtime loaded. If the earliest runtime is 3.1 then you wouldn't see this error because the version of the System.Runtime is 4.2.2.0.

The Proposed Solution:
Build against netcoreapp3.1 and require a suitable runtime to satisfy the .NET 5 SDK Build.Tasks requirement. It seems likely that more users of dotnet-format will have .NET 5 SDKs installed and start seeing this error themselves.


JoeRobich:
Closing this for now, since we enabled Runtime rollforward.

Issue #670, opened 2020-05-11
Allow running on generated code


sharwell:
**dotnet-format** should have a command line flag that allows the formatter to run on files that would otherwise be considered generated. This will allow the tool to be used by the code generation tools themselves to produce properly-formatted code.

jmarolf:
You should be able to do this today with the `--files` option

JoeRobich:
@jmarolf We will need to work around the generated code check.

jmarolf:
ah, got it

PR #673, opened 2020-05-12
Add option to format generated code files.

JoeRobich:
Resolves https://github.com/dotnet/format/issues/670

### DIFF
--- a/src/Program.cs
+++ b/src/Program.cs
@@ -55,7 +55,7 @@ namespace Microsoft.CodeAnalysis.Tools
                 {
                     Argument = new Argument<string?>() { Arity = ArgumentArity.ExactlyOne }
                 },
-                new Option(new[] { "--include-generated" }, Resources.Inlcude_generated_code_files_in_formatting_operations)
+                new Option(new[] { "--include-generated" }, Resources.Include_generated_code_files_in_formatting_operations)
                 {
                     Argument = new Argument<bool>(),
                     IsHidden = true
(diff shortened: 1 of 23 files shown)


src/Program.cs
```diff
@@ -55,6 +55,11 @@ private static async Task<int> Main(string[] args)
                 {
                     Argument = new Argument<string?>() { Arity = ArgumentArity.ExactlyOne }
                 },
+                new Option(new[] { "--include-generated" }, Resources.Inlcude_generated_code_files_in_formatting_operations)
+                {
+                    Argument = new Argument<bool>(),
+                    IsHidden = true
```
JoeRobich, line 61: I marked it as hidden because I don't expect this to be super common and I didn't want to add more to the help screen. I can likely be talked down from this stance.

src/Program.cs
```diff
@@ -55,6 +55,11 @@ private static async Task<int> Main(string[] args)
                 {
                     Argument = new Argument<string?>() { Arity = ArgumentArity.ExactlyOne }
                 },
+                new Option(new[] { "--include-generated" }, Resources.Inlcude_generated_code_files_in_formatting_operations)
```
sharwell, line 58:
```suggestion
                new Option(new[] { "--include-generated" }, Resources.Include_generated_code_files_in_formatting_operations)
```

JoeRobich: doh. thanks

PR #685, opened 2020-05-20
Add using directive formatting

wojciechcibor:
I was playing around with dotnet-format and just for fun added support for following .editorconfig flags:

dotnet_sort_system_directives_first
dotnet_separate_import_directive_groups

It works for me. Please let me know what do you think about it. There is one TODO about using warning from Resources.

### DIFF
--- a/docs/Supported-.editorconfig-options.md
+++ b/docs/Supported-.editorconfig-options.md
@@ -60,3 +60,9 @@ These formatting rules concern the use of single lines versus separate lines for
 
 - csharp_preserve_single_line_statements (default value: `true`)
 - csharp_preserve_single_line_blocks (default value: `true`)
+
+**Organize using directives**
+These formatting rules concern the sorting and display of using directives and Imports statements.
+
+- dotnet_sort_system_directives_first  (default value: `true`)
+- dotnet_separate_import_directive_groups  (default value: `true`)
--- a/src/CodeFormatter.cs
+++ b/src/CodeFormatter.cs
@@ -29,6 +29,7 @@ namespace Microsoft.CodeAnalysis.Tools
             new FinalNewlineFormatter(),
             new EndOfLineFormatter(),
             new CharsetFormatter(),
+            new ImportsFormatter(),
         }.ToImmutableArray();
 
         public static async Task<WorkspaceFormatResult> FormatWorkspaceAsync(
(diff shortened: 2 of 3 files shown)


JoeRobich:
@wojciechcibor Thanks for opening this PR! It has been on our todo list for a while and we appreciate your contribution. 

For this to move forward, we would need to act on the todo and move the fix description to our .resx. We would also need to add some simple tests to ensure the formatter was being invoked. If you have the time and feel up to implementing these, I am happy to give guidance. Otherwise, I could make these final changes. How would you like to proceed?

wojciechcibor:
@JoeRobich Thanks for reply. 

Well, at first just wanted to see If this has any value for you - before that I didn't want go to much into it.
In this case I can work futher on that. I can implement this TODO and add/fix tests - perhaps even later on today.

JoeRobich:
@wojciechcibor That's great! The tests don't have to be long or involved since the Formatter is already tested in dotnet/roslyn. You can use the tests from the feature/analyzers branch - https://github.com/dotnet/format/blob/feature/analyzers/tests/Formatters/ImportsFormatterTests.cs

JoeRobich:
@wojciechcibor We are at the point that we would like to do a 4.x release and this would be a great feature to include. Would you have time to fix up this PR in the next day or two?

JoeRobich:
@wojciechcibor Let me apologize for saying adding the tests would be easy. I gave it a go yesterday afternoon and now realize, with the move to the Roslyn .editorconfig support, we aren't adding an AnalyzerConfigDocument to the Workspace, which means the document options are always default. If you don't mind I would like to fix this myself and we will get your changes merged in.

JoeRobich:
Thanks @wojciechcibor this will be in the next release!

virzak:
Is there a way to skip this just this one rule?

JoeRobich:
@virzak We will be shipping a 4.1 release this week that has a change to only run this formatter when either of the driving properties are configured in the .editorconfig (https://github.com/dotnet/format/pull/701). Will this work in your scenario?

virzak:
Yeah, certainly. My azure-pipelines.yml always gets the latest version of dotnet-format (which is probably a bad idea to begin with on my part). The latest version forces to change almost all files. I'll target the build to version 3.3.111304 to make sure that the old builds aren't failing and that a switch to a newer version is controlled.

PR #689, opened 2020-05-25
Added option checks to warn about deprecations

JoeRobich:
- Moves the root command definition and validators to a separate class.
- Hides the `--workspace` option.
- Deprecates specifying the folder path via the `--folder` option. Warning is shown to specify the path via the `<project>` argument and to simply use `--folder` as a flag.
- Warning is shown when using `--dry-run` instead of `--check`.
- Warning is shown when using `--files` instead of `--include`.

The plan would be to remove the warnings and alias options after the 4.x release.

### DIFF
--- a/src/Program.cs
+++ b/src/Program.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.CommandLine;
 using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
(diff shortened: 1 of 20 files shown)


JoeRobich:
@jmarolf Please take a look

src/Program.cs
```diff
@@ -27,112 +26,29 @@ internal class Program
...
+
+            // Parse the incoming args so we can give warnings when deprecated options are used.
+            s_parseResult = rootCommand.Parse(args);
 
-            // Parse the incoming args and invoke the handler
             return await rootCommand.InvokeAsync(args);
         }
 
-        internal static RootCommand CreateCommandLineOptions()
+        public static async Task<int> Run(string? project, string? folder, string? workspace, string? verbosity, bool check, string[] include, string[] exclude, string? report, bool includeGenerated, IConsole console = null!)
```
jmarolf, line 42: It may finally be time to wrap this line

JoeRobich: Curses!

I'll get it done.

PR #693, opened 2020-05-29
Add Imports Formatter

JoeRobich:
Builds on the work from https://github.com/dotnet/format/pull/685.  Thanks @wojciechcibor!

- Adds AnalyzerConfigDocument to test workspace, so that AnalyzerConfigOptions are flowing into DocumentOptionSet properly.
- Adds resource string for ImportsFormatter fix message
- Adds ImportsFormatterTests from features/analyzers


JoeRobich:
Opened https://github.com/dotnet/roslyn/issues/44710 for the Roslyn formatting error

src/Formatters/ImportsFormatter.cs
```diff
@@ -0,0 +1,53 @@
...
+            OptionSet optionSet,
+            AnalyzerConfigOptions? analyzerConfigOptions,
+            FormatOptions formatOptions,
+            ILogger logger,
+            CancellationToken cancellationToken)
+        {
+            try
+            {
+                var organizedDocument = await Formatter.OrganizeImportsAsync(document, cancellationToken);
+                if (organizedDocument == document)
```
jmarolf, line 34: I didn't think equality for documents was implemented. Did we add that? I would normally compare the document ids.

src/Formatters/ImportsFormatter.cs
```diff
@@ -0,0 +1,53 @@
...
+
+                // Because the Formatter does not abide the `end_of_line` option we have to fix up the ends of the organized lines.
+                // See https://github.com/dotnet/roslyn/issues/44136
+                var organizedSourceText = await organizedDocument.GetTextAsync(cancellationToken);
+                return await _endOfLineFormatter.FormatFileAsync(organizedDocument, organizedSourceText, optionSet, analyzerConfigOptions, formatOptions, logger, cancellationToken);
+            }
+            catch (InsufficientExecutionStackException)
+            {
+                // This case is normally not hit when running against a handwritten code file.
+                // https://github.com/dotnet/roslyn/issues/44710#issuecomment-636253053
```
jmarolf, line 47: We should probably tell the user that the file was too large to format.

tests/Formatters/AbstractFormatterTests.cs
```diff
@@ -50,21 +49,25 @@ static AbstractFormatterTest()
...
         }
 
-        protected virtual string DefaultFilePathPrefix { get; } = "Test";
+        protected virtual string DefaultFilePathPrefix => "Test";
 
-        protected virtual string DefaultTestProjectName { get; } = "TestProject";
+        protected virtual string DefaultTestProjectName => "TestProject";
 
-        protected virtual string DefaultTestProjectPath => "." + Path.DirectorySeparatorChar + DefaultTestProjectName + "." + DefaultFileExt + "proj";
+        protected virtual string DefaultFolderPath => Path.DirectorySeparatorChar + DefaultTestProjectName + Path.DirectorySeparatorChar;
```
jmarolf, line 56: Can we `Path.Combine` for all of these cases? Why append the slashes manually?

src/Formatters/ImportsFormatter.cs
JoeRobich: Well, I just want to know if the document was changed, so I thought a ref equals would be fine. Could use `HasTextChanged` instead tho.

src/Formatters/ImportsFormatter.cs
JoeRobich: Yeah, although is it a size issue or an expression complexity issue. Perhaps can just be vague and say "Unable to organize file `{filename}`"

tests/Formatters/AbstractFormatterTests.cs
JoeRobich: The beginning separator is important because I need the path to appear rooted. Using Path.Combine elsewhere is a thing i can do tho.

src/Formatters/ImportsFormatter.cs
jmarolf: I believe the message the compiler gives here is `expression is too complex`

src/Formatters/ImportsFormatter.cs
jmarolf: yeah, `HasTextChanged` seems the most efficient implementation since it looks at the text version

JoeRobich: And it is internal. =) I could run GetTextChangesAsync and check to see if it is empty.

JoeRobich: Actually, I've verified the ref check is doing exactly what I want it to do. The call to Organize is returning the same document instance when there are no changes.

jmarolf: ah, that's why I always recommend customers compare `DocumentId`s, I forgot. `DocumentId` implements equitable so you can do `(organizedDocument.Id == document.Id)`

jmarolf: > The call to Organize is returning the same document instance when there are no changes.

Good to know. Feel free to ignore this then. My only nit is that I think this behavior is an implementation detail and not part of the public API contract.

PR #694, opened 2020-05-30
Always run on the latest Runtime

JoeRobich:
As an alternative fix to the move to netcoreapp3.1, we could always run on the latest installed runtime. We were already rolling forward in cases when the requested runtime was not installed.


JoeRobich:
@sharwell @jmarolf Thoughts on this approach?

sharwell:
I like this approach

eng/format-verifier.ps1
```diff
@@ -12,6 +12,11 @@ if (!(Test-Path $testPath)) {
     New-Item -ItemType Directory -Force -Path $testPath | Out-Null
 }
 
+if (Test-Path '.\eng\common\Build.ps1') {
```
sharwell, line 15: 😕 Can this ever fail?

JoeRobich: nope =)

JoeRobich:
There was an issue running on 3.0.1xx preview runtimes that did not allow 2.x apps to rollforward. Fixed in https://github.com/dotnet/sdk/pull/3542

tests/dotnet-format.UnitTests.csproj
```diff
@@ -1,10 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
RikkiGibson, line 4: did we mean to roll this back?

JoeRobich: Yeah, since we are rolling forward I wanted running the unit tests to behave the same a running dotnet-format.

jmarolf: the roll forward _should_ allow this to work without issue

Issue #691, opened 2020-05-27
Refactor out the DocumentWithOptions type


JoeRobich:
Based on the suggestion https://github.com/dotnet/format/pull/590#discussion_r430790777

We can load options to determine formattable files then pass around a array of DocumentId. DocumentOptions and AnalyzerOptionSet are cached and we can retrieve them again before running the formatters.

PR #695, opened 2020-06-01
Remove DocumentWithOptions and pass around DocumentIds

JoeRobich:
Resolves https://github.com/dotnet/format/issues/691


JoeRobich:
Anecdotally this improves performance

Integration Run Before Changes
image

Integration Run With Changes
image

tiesmaster:
@JoeRobich As a lurker in this repo, I saw this PR coming by, and I was just curious why the integration steps take this long, so I was looking into the integration tests, and what takes up its time. Opening up [`azure-pipelines-integration.yml`](https://github.com/dotnet/format/blob/master/), I noticed that both this repo, as Roslyn have [the same hash reference](https://github.com/dotnet/format/blob/master/azure-pipelines-integration.yml#L22-L26) (which of course cannot happen with git).

Looking at the pipeline task, I also see [it fails](https://dev.azure.com/dnceng/public/_build/results?buildId=667106&view=logs&j=41fd5339-8f2d-502b-c114-8afb1cf469d4&t=d1ead47c-c028-59e7-76db-a09be00e642c&l=112) on the checkout, so you might want to look at that.

JoeRobich:
@ tiesmaster Thanks for lurking!

> I was just curious why the integration steps take this long

image

Roslyn is definitely the long pole in the integration tests. 

The prepare step which takes over a third of the time, is where we clone the repo, checkout the SHA we want to target, and restore all the solutions we discover. It is possible we can improve on the cloning by only checking out the SHA we care about with a depth of 1 (see https://stackoverflow.com/a/43136160).

The second longest step is Running format against the Roslyn.sln. This can be painful because there are, at last count, over 170 projects referenced. So it takes a while for the design time builds to complete so that we know which files are part of the solution. Loading fully from a design time build also provides us with the semantic information we will need to run analyzers and fixes in the future. In addition we also run against the Compiler.sln, so a double whammy there.

>  I noticed that both this repo, as Roslyn have the same hash reference (which of course cannot happen with git).

Thanks for spotting this. Cut and paste strikes again =)

PR #711, opened 2020-06-29
use explicit types for commandline parsing

jmarolf:


src/FormatCommand.cs
```diff
@@ -58,11 +62,84 @@ internal static RootCommand CreateCommandLineOptions()
...
+            {
+                // Use Defaults
+                return DiagnosticSeverity.Error;
+            }
+
+            var severityValue = result.Tokens.Single().Value;
+            if (!HasMatch(severityValue, out var severity))
+            {
+                // Assume that this means our arity is zero and the parser should re-parse accordingly
+                result.ErrorMessage = $"Cannot specify severity more than once for {result.Argument.Aliases.FirstOrDefault()}";
```
JoeRobich, line 113: Error message not applicable

src/FormatCommand.cs
```diff
@@ -58,11 +62,84 @@ internal static RootCommand CreateCommandLineOptions()
...
+            {
+                result.ErrorMessage = "Cannot specify verbosity more than once.";
+                return null;
+            }
+
+            if (result.Tokens.Count == 0)
+            {
+                result.ErrorMessage = "Must specify a verbostiy level.";
+                return null;
+            }
```
JoeRobich, line 77: Are these checks necessary since we mark the argument arity as ExactlyOne?

src/FormatCommand.cs
JoeRobich: Is this possible since we mark the argument with valid values using FromAmong?

src/FormatCommand.cs
```diff
@@ -58,11 +62,84 @@ internal static RootCommand CreateCommandLineOptions()
...
+        private static DiagnosticSeverity? ParseDiagnosticSeverity(ArgumentResult result)
         {
-            return result.Tokens
-                .Where(token => token.Type == TokenType.Option)
-                .Any(token => aliases.Contains(token.Value));
+            if (result.Tokens.Count > 1)
+            {
+                result.ErrorMessage = $"Cannot specify severity more than once for {result.Argument.Name}";
+                return null;
+            }
```
JoeRobich, line 101: Is this necessary since we mark the argument arity as ZeroOrOne

src/FormatCommand.cs
jmarolf: blocked by https://github.com/dotnet/command-line-api/pull/955

jmarolf:
This would require us to write a completely custom help presenter. I think everything should stay stringly-typed for now

PR #713, opened 2020-07-02
Merge in Feature/analyzers

jmarolf:
I think the other cleanup issues should not block merging this


JoeRobich:
updated msftbot to ignore the checks that had blocked it.

PR #720, opened 2020-07-02
Run naming style analyzer

JoeRobich:

PR #723, opened 2020-07-06
Run all analyzers even when a fixer isn't present

JoeRobich:
This builds on the work and refactoring done in https://github.com/dotnet/format/pull/720

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -131,6 +131,7 @@ csharp_indent_switch_labels = true
 csharp_indent_labels = flush_left
 
 # Prefer "var" everywhere
+dotnet_diagnostic.IDE0007.severity = error
 csharp_style_var_for_built_in_types = true:error
 csharp_style_var_when_type_is_apparent = true:error
 csharp_style_var_elsewhere = true:error
@@ -178,3 +179,35 @@ csharp_space_between_parentheses = false
 csharp_prefer_braces = true:silent
 csharp_preserve_single_line_blocks = true
 csharp_preserve_single_line_statements = true
+
+# IDE0073: File header
+dotnet_diagnostic.IDE0073.severity = warning
+file_header_template = Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
+
+# IDE0035: Remove unreachable code
+dotnet_diagnostic.IDE0035.severity = warning
+
+# IDE0036: Order modifiers
+dotnet_diagnostic.IDE0036.severity = warning
+
+# IDE0040: Add accessibility modifiers
+dotnet_diagnostic.IDE0040.severity = warning
+
+# IDE0043: Format string contains invalid placeholder
+dotnet_diagnostic.IDE0043.severity = warning
+
+# IDE0044: Make field readonly
+dotnet_diagnostic.IDE0044.severity = warning
+
+# CONSIDER: Are IDE0051 and IDE0052 too noisy to be warnings for IDE editing scenarios? Should they be made build-only warnings?
+# IDE0051: Remove unused private member
+dotnet_diagnostic.IDE0051.severity = warning
+
+# IDE0052: Remove unread private member
+dotnet_diagnostic.IDE0052.severity = warning
+
+# IDE0059: Unnecessary assignment to a value
+dotnet_diagnostic.IDE0059.severity = warning
+
+# IDE0060: Remove unused parameter
+dotnet_diagnostic.IDE0060.severity = warning
--- a/src/Analyzers/AnalyzerFormatter.cs
+++ b/src/Analyzers/AnalyzerFormatter.cs
@@ -17,18 +17,18 @@ namespace Microsoft.CodeAnalysis.Tools.Analyzers
     internal class AnalyzerFormatter : ICodeFormatter
     {
         private readonly string _name;
-        private readonly IAnalyzerFinder _finder;
+        private readonly IAnalyzerInformationProvider _informationProvider;
         private readonly IAnalyzerRunner _runner;
         private readonly ICodeFixApplier _applier;
 
         public AnalyzerFormatter(
             string name,
-            IAnalyzerFinder finder,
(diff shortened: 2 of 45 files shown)


.editorconfig
```diff
@@ -178,3 +179,35 @@ csharp_space_between_parentheses = false
 csharp_prefer_braces = true:silent
 csharp_preserve_single_line_blocks = true
 csharp_preserve_single_line_statements = true
+
+# IDE0073: File header
+dotnet_diagnostic.IDE0073.severity = warning
+file_header_template = Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
```
JoeRobich, line 185: Added some of the goodness from Roslyn's .editorconfig

src/Analyzers/AnalyzerOptionExtensions.cs
```diff
@@ -44,14 +42,17 @@ private static string GetCategoryBasedDotnetAnalyzerDiagnosticSeverityKey(string
...
+            if (compilation.Options.SpecificDiagnosticOptions.TryGetValue(descriptor.Id, out severity))
+            {
+                return true;
+            }
+
+            // If user has explicitly configured severity for this diagnostic ID, that should be respected.
             // For example, 'dotnet_diagnostic.CA1000.severity = error'
-            if (compilation.Options.SpecificDiagnosticOptions.ContainsKey(descriptor.Id) ||
-                tree.DiagnosticOptions.ContainsKey(descriptor.Id))
+            if (tree.DiagnosticOptions.TryGetValue(descriptor.Id, out severity))
```
JoeRobich, line 53: We were not returning the severity from DiagnosticOptions causing some analyzers to not be run.

src/Analyzers/AnalyzerRunner.cs
```diff
@@ -11,20 +11,29 @@ namespace Microsoft.CodeAnalysis.Tools.Analyzers
 {
     internal partial class AnalyzerRunner : IAnalyzerRunner
     {
+        private readonly bool _includeComplilerDiagnostics;
+
+        public AnalyzerRunner(bool includeCompilerDiagnostics)
+        {
+            _includeComplilerDiagnostics = includeCompilerDiagnostics;
```
JoeRobich, line 18: This should open up the possibility of running fixers that have no analyzer because they work against compiler diagnostics.

src/CodeFormatter.cs
```diff
@@ -29,8 +29,8 @@ internal static class CodeFormatter
             new EndOfLineFormatter(),
             new CharsetFormatter(),
             new ImportsFormatter(),
-            new AnalyzerFormatter(Resources.Code_Style, new RoslynCodeStyleAnalyzerFinder(), new AnalyzerRunner(), new SolutionCodeFixApplier()),
-            new AnalyzerFormatter(Resources.Analyzer_Reference, new AnalyzerReferenceAnalyzerFinder(), new AnalyzerRunner(), new SolutionCodeFixApplier()),
+            new AnalyzerFormatter(Resources.Code_Style, new CodeStyleInformationProvider(), new AnalyzerRunner(includeCompilerDiagnostics: true), new SolutionCodeFixApplier()),
+            new AnalyzerFormatter(Resources.Analyzer_Reference, new AnalyzerReferenceInformationProvider(), new AnalyzerRunner(includeCompilerDiagnostics: false), new SolutionCodeFixApplier()),
```
JoeRobich, line 33: Possibly we should always be including compiler diagnostics.

src/Analyzers/AnalyzerFormatter.cs
```diff
@@ -130,40 +132,121 @@ static void LogDiagnosticLocations(Solution solution, IEnumerable<Diagnostic> di
...
+                        }
                     }
                 }
             }
 
             return solution;
+
+            static ImmutableDictionary<string, ImmutableArray<DiagnosticAnalyzer>> CreateAnalyzerMap(
+                ImmutableArray<string> diagnosticIds,
+                ImmutableArray<DiagnosticAnalyzer> analyzers)
```
jmarolf, line 198: looks like we use an inconsistent wrapping style. Which one do we prefer?

src/Analyzers/AnalyzerFormatter.cs
```diff
@@ -130,40 +132,121 @@ static void LogDiagnosticLocations(Solution solution, IEnumerable<Diagnostic> di
...
+            }
+
+            // Build maps between diagnostic id and the associated analyzers and codefixes
+            var analyzersById = CreateAnalyzerMap(reportedDiagnostics, allAnalyzers);
+            var fixersById = CreateFixerMap(reportedDiagnostics, allCodefixes);
+
+            // We need to run each codefix iteratively so ensure that all diagnostics are found and fixed.
+            foreach (var diagnosticId in reportedDiagnostics)
             {
+                var analyzers = analyzersById[diagnosticId];
```
jmarolf, line 157: perhaps we should try to get the value from the dictionary? Isn't it expected that we would not find entries for this?

jmarolf: Ah I see, the `ToImmutableArray` on an empty enumerable will mean that we should always have an empty immutable array

---
In reply to: [450373591](https://github.com/dotnet/format/pull/723#discussion_r450373591) [](ancestors = 450373591)

src/Analyzers/AnalyzerOptionExtensions.cs
jmarolf: Well, this is why we need a test for this :(

src/CodeFormatter.cs
jmarolf: yeah, I think we should play it safe and always generate compiler diagnostics.

src/Analyzers/AnalyzerFormatter.cs
JoeRobich: I like this one because it doesn't change when things are renamed/refactored.

PR #725, opened 2020-07-06
Filter analyzers by project language before running

JoeRobich:


jmarolf:
I need to write tests for this

src/Analyzers/AnalyzerFormatter.cs
```diff
@@ -193,13 +200,18 @@ static void LogDiagnosticLocations(Solution solution, IEnumerable<Diagnostic> di
 
             return solution;
 
-            static ImmutableDictionary<string, ImmutableArray<DiagnosticAnalyzer>> CreateAnalyzerMap(
+            static ImmutableDictionary<string, ImmutableDictionary<string, ImmutableArray<DiagnosticAnalyzer>>> CreateAnalyzerMap(
```
RikkiGibson, line 203: I get it, but also this dictionary will have max two entries that you care about, there are no efforts in the works to generalize roslyn analyzers/fixers to more languages, consider just making the dictionary value a pair of arrays for CS and VB respectively.

jmarolf: look at mr. glass-half-empty thinking we won't support hundreds of lanuguages

Issue #718, opened 2020-07-02
Add analyzer fixing to integration tests


jmarolf:

JoeRobich:
At the moment integration tests run dotnet-format against all solution files in the first 2 levels of repo folders. Perhaps we should limit this to a single known solution file per repo and enable running analyzers to compensate for the additional processing time.

PR #728, opened 2020-07-08
Run analyzers during integration tests

JoeRobich:
resolves https://github.com/dotnet/format/issues/718


src/dotnet-format.csproj
```diff
@@ -17,6 +17,7 @@
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
 
     <!-- Always run on the latest runtime installed. -->
```
jmarolf, line 19: I don't know that the comment should say to explain why it needs to be `rc1` but we should do a favor to ourselves and write that down

JoeRobich:
I imagine this would complete well before 33 minutes if the netstandard references were being added to the projects...

JoeRobich:
With references loading, the run time for Roslyn repo was 27 minutes.

Issue #734, opened 2020-07-17
fix-analyzers throwing error.


ChrisBertrand:
When running the latest development build.  --version 5.0.136601

It runs through until it hits, Running Formatters then throws an error. Fix-style is working fine.

Running formatters.
Unhandled exception: System.IO.FileLoadException: Assembly with same name is already loaded
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromPath(IntPtr ptrNativeAssemblyLoadContext, String ilPath, String niPath, ObjectHandleOnStack retAssembly)
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(String assemblyPath)
   at System.Reflection.Assembly.LoadFrom(String assemblyFile)
   at Microsoft.CodeAnalysis.Tools.Analyzers.AnalyzerReferenceInformationProvider.<>c.<GetAnalyzersAndFixers>b__0_1(String path) in /_/src/Analyzers/AnalyzerReferenceInformationProvider.cs:line 28

The full set of analyzers returned are: 
```
[0]: ".nuget\\packages\\microsoft.entityframeworkcore.analyzers\\3.1.4\\analyzers\\dotnet\\cs\\Microsoft.EntityFrameworkCore.Analyzers.dll"
    [1]: "packages\\Microsoft.CodeAnalysis.VersionCheckAnalyzer.2.9.6\\analyzers\\dotnet\\cs\\Microsoft.CodeAnalysis.VersionCheckAnalyzer.resources.dll"
    [2]: "packages\\Microsoft.CodeAnalysis.VersionCheckAnalyzer.2.9.6\\analyzers\\dotnet\\Microsoft.CodeAnalysis.VersionCheckAnalyzer.dll"
    [3]: "packages\\Microsoft.CodeQuality.Analyzers.2.9.6\\analyzers\\dotnet\\cs\\Humanizer.dll"
    [4]: "packages\\Microsoft.CodeQuality.Analyzers.2.9.6\\analyzers\\dotnet\\cs\\Microsoft.CodeQuality.Analyzers.dll"
    [5]: "packages\\Microsoft.CodeQuality.Analyzers.2.9.6\\analyzers\\dotnet\\cs\\Microsoft.CodeQuality.CSharp.Analyzers.dll"
    [6]: "packages\\Microsoft.NetCore.Analyzers.2.9.6\\analyzers\\dotnet\\cs\\Microsoft.NetCore.Analyzers.dll"
[...]

PR #746, opened 2020-08-06
Load analyzer assemlbies in their own AssemblyLoadContext

JoeRobich:
Attempt to address https://github.com/dotnet/format/issues/734


src/Analyzers/AnalyzerReferenceInformationProvider.cs
```diff
@@ -1,35 +1,94 @@
...
+                    if (File.Exists(possibleAssemblyPath))
+                    {
+                        return LoadFromAssemblyPath(possibleAssemblyPath);
+                    }
+                }
+                catch { }
+
+                // Try to load the requested assembly from the default load context.
+                return AssemblyLoadContext.Default.LoadFromAssemblyName(assemblyName);
+            }
```
JoeRobich, line 91: This nees to be better. =)

jmarolf: its a great start!

JoeRobich: Updated to use the same logic as the LooseAssemblyVersionLoader

jmarolf (commented):
:)

Issue #398, opened 2019-10-18
Run compiler diagnostics if any active code fix declares support for them


sharwell:
If any active code fix contains a CS* (C#) or BC* (Visual Basic) diagnostic ID in the set of supported diagnostics, we need to make sure these compiler diagnostics are gathered.

Filed as a follow-up to #281

jmarolf:
right, we will. my understanding is that today no code style warning that we ship needs this. We'll also need to investigate how poorly this effects performance

sharwell:
I believe we have some code fixes in place for fixing some compiler warnings. It should be a straightforward change to check the declared diagnostic IDs though to make sure we are ready for a case if/when a user adds a code fix to the mix that uses them (e.g. a code fix that adds NRT annotations to explicitly-typed local variables).

PR #750, opened 2020-08-09
Only include compiler diagnostics if a fixer supports them

JoeRobich:
implements https://github.com/dotnet/format/issues/398

### DIFF
--- a/src/Analyzers/AnalyzerRunner.cs
+++ b/src/Analyzers/AnalyzerRunner.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
 
 using System.Collections.Immutable;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.Diagnostics;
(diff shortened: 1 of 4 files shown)


src/Analyzers/AnalyzerRunner.cs
```diff
@@ -45,7 +39,7 @@ public AnalyzerRunner(bool includeCompilerDiagnostics)
 
             // If are not running any analyzers and are not reporting compiler diagnostics, then there is
             // nothing to report.
-            if (analyzers.IsEmpty && !_includeComplilerDiagnostics)
+            if (analyzers.IsEmpty && !includeCompilerDiagnostics)
```
RikkiGibson, line 42: it feels like we could save on the boilerplate of passing this down if we just compute 'includeCompilerDiagnostics' here from the list of analyzers.

RikkiGibson: perhaps there are tests which need to directly specify this?

JoeRobich: it is calculated from the list of codefixproviders.

JoeRobich: If a test is testing this method directly, they could pass this value in as desired. If they were testing this indirectly, they would need to build and add an appropriate AnalyzerReference containing a fixer for a compiler diagnostic.

RikkiGibson: Oops, my mistake!

Issue #744, opened 2020-08-04
Feature request: Port VS's Organize Usings facility to dotnet-format


am11:
This feature request is about organizing `using namespace` from command-line tool. This feature is often used in Visual Studio in context of 'code formatting'.

Visual Studio:

![image](https://user-images.githubusercontent.com/3840695/89336409-03803000-d6a2-11ea-9997-2e613825fc50.png)
![image](https://user-images.githubusercontent.com/3840695/89336088-881e7e80-d6a1-11ea-91cc-bdec7944f6d7.png)

Desired Options from dotnet-format:

```diff
dotnet-format:
  dotnet-format

Usage:
  dotnet-format [options] [<project>]

Arguments:
  <project>    The solution or project file to operate on. If a file is not specified, the command will search the current directory for one.

Options:
  -f, --folder <folder>           Whether to treat the `<project>` path as a folder of files. [default: ]
  --files, --include <include>    A list of relative file or folder paths to include in formatting. All files are formatted if empty. [default: System.String[]]
  --exclude <exclude>             A list of relative file or folder paths to exclude from formatting. [default: System.String[]]
  --check, --dry-run              Formats files without saving changes to disk. Terminates with a non-zero exit code if any files were formatted.
  --report <report>               Accepts a file path, which if provided, will produce a json report in the given directory. [default: ]
+  -r, --remove-usings            Removes unnecessary usings.
+  -s, --sort-usings[=system]     Sorts usings alphabetically. With value 'system', places 'System' directives first.
  -v, --verbosity <verbosity>     Set the verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic]
  --version                       Show version information
  -?, -h, --help                  Show help and usage information
```

JoeRobich:
@am11 We have support for organizing imports using the following .editorconfig settings - https://docs.microsoft.com/en-us/visualstudio/ide/editorconfig-formatting-conventions?view=vs-2019#organize-using-directives

To support removing unused imports, I was hoping that there would be a working fixer for IDE0005. However, we can conditionally run `document.GetLanguageService<IRemoveUnnecessaryImportsService>().RemoveUnnecessaryImportsAsync(document, cancellationToken)` as part of `--fix-style` if we detect IDE0005 is configured.

mgedda:
> 
> 
[...]

PR #753, opened 2020-08-11
Remove aliases for the fix style and fix analyzers options

JoeRobich:
Based on https://github.com/dotnet/format/issues/744#issuecomment-671962627 as well as the difficulty we had when the `-f` option accepted an argument, we are not using aliases correctly. They are intended to be a single letter. Let's remove these until we have a better plan for them.

### DIFF
--- a/src/FormatCommand.cs
+++ b/src/FormatCommand.cs
@@ -22,11 +22,11 @@ namespace Microsoft.CodeAnalysis.Tools
                     Description = Resources.A_path_to_a_solution_file_a_project_file_or_a_folder_containing_a_solution_or_project_file_If_a_path_is_not_specified_then_the_current_directory_is_used
                 }.LegalFilePathsOnly(),
                 new Option(new[] { "--folder", "-f" }, Resources.Whether_to_treat_the_workspace_argument_as_a_simple_folder_of_files),
-                new Option(new[] { "--fix-style", "-fs" }, Resources.Run_code_style_analyzers_and_apply_fixes)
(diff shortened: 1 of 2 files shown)


src/FormatCommand.cs
```diff
@@ -22,11 +22,11 @@ internal static RootCommand CreateCommandLineOptions()
                     Description = Resources.A_path_to_a_solution_file_a_project_file_or_a_folder_containing_a_solution_or_project_file_If_a_path_is_not_specified_then_the_current_directory_is_used
                 }.LegalFilePathsOnly(),
                 new Option(new[] { "--folder", "-f" }, Resources.Whether_to_treat_the_workspace_argument_as_a_simple_folder_of_files),
-                new Option(new[] { "--fix-style", "-fs" }, Resources.Run_code_style_analyzers_and_apply_fixes)
+                new Option(new[] { "--fix-style" }, Resources.Run_code_style_analyzers_and_apply_fixes)
```
am11, line 25: Would it make sense if we change it to `-s` and other one `-a`?
nit: readme.md file will need an update. 🙂

jmarolf: I agree that we should change all the short-cut args to be single letter. This also ensures that we don't accidentally alias another options. `-s` and `-a` seem reasonable to me as short hand for `--fix-style` and `--fix-analyzers`

JoeRobich: It might. I think removing them for a time would be a good thing, because, as you point out, if users are using `-fs` or `-fa` then those will be interpreted as `--folder` and one of the fix options.

Issue #757, opened 2020-08-14
Slow performance for --include with single file


chriselion:
Hi,
I'm just experimenting with dotnet-format for the first time. I noticed that running it on a single file has more overhead than I would expect. 

```
# Run on 223 files 
$ git ls-files | grep .cs$  > all_cs.txt
$ dotnet format --folder --include @all_cs.txt
  Formatting code files in workspace '/xyz/ml-agents'.
  ...
  Format complete in 2297ms.

# Run on 1 file
$ dotnet format --folder --include com.unity.ml-agents/Tests/Editor/MultinomialTest.cs
  Formatting code files in workspace '/xyz/ml-agents'.
  Format complete in 1325ms.
```
Is that expected? It's not horrible, but it would be mildly annoying to run incrementally on git commits.

Platform/version information:
MacOS 10.14.6
```
$ dotnet --version
3.1.401
$ dotnet-format --version
4.1.131201+82a5527cd93d66bffc885c0bb901499f5981db35
```

RikkiGibson:
I am not surprised to hear that performance suffers in a project with a single large file compared to a project with similar source code split across multiple files. We only use a single thread to parse each syntax tree (IIRC). I am not deeply familiar with Roslyn's formatting code but I imagine it has similar limitations, where it can't fully utilize your CPU cores unless you have enough files for them all to run in parallel.

RikkiGibson:
(sorry, I misunderstood the description in my original comment.) It does feel like there's an excessive amount of fixed-size overhead here.

chriselion:
Thanks, sorry if I was being unclear - the fixed-size overhead was what I was complaining about.

Any profiling information that I can provide to help diagnose? I haven't tried `diag` verbosity yet; would there be anything useful in there?

RikkiGibson:
A trace from a sampling profiler of some kind could be helpful. My guess though is that a lot of this overhead is in just spinning up the runtime, loading the compiler assemblies and jitting methods--assuming that we do not use any kind of multiproc or client/server model like we do when you run the compiler in order to build. Do you know @JoeRobich @jmarolf?

jmarolf:
would be helpful to know if there is any better perf in the latest (unrelesed) version.

```
dotnet tool install -g dotnet-format --version 5.0.135301 --add-source https://dotnet.myget.org/F/format/api/v3/index.json
```

[...]

PR #763, opened 2020-08-16
Improve folder workspace performance

JoeRobich:
Improve https://github.com/dotnet/format/issues/757. This PR does a few things. 

- Removes Linq from the FolderWorkspace classes
- Special case when --include is only a list of filenames (no globs or folders)
- Extracts report writing out of the CodeFormatter class

### DIFF
--- a/src/CodeFormatter.cs
+++ b/src/CodeFormatter.cs
@@ -42,7 +42,7 @@ namespace Microsoft.CodeAnalysis.Tools
             var workspaceStopwatch = Stopwatch.StartNew();
 
             using var workspace = formatOptions.WorkspaceType == WorkspaceType.Folder
-                ? OpenFolderWorkspaceAsync(formatOptions.WorkspaceFilePath, formatOptions.FileMatcher)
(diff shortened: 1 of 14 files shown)


src/CodeFormatter.cs
```diff
@@ -131,31 +111,11 @@ internal static class CodeFormatter
...
-                return Path.Combine(Environment.CurrentDirectory, defaultReportName);
-            }
-            else
-            {
-                return Path.Combine(reportPath, defaultReportName);
-            }
-        }
-
-        private static Task<Workspace> OpenFolderWorkspaceAsync(string workspacePath, Matcher fileMatcher, CancellationToken cancellationToken)
+        private static Workspace OpenFolderWorkspaceAsync(string workspacePath, SourceFileMatcher fileMatcher)
```
jmarolf, line 114: rename to `OpenFolderWorkspace` as it doesn't return a task

JoeRobich: haha. This is why I need a second set of eyes =)

PR #767, opened 2020-08-20
Fix warnings and apply suggestions.

JoeRobich:
I locally updated the TFM to net5.0. I fixed the reported nullability warnings and applied the analyzer suggestions that were offered.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -11,7 +11,7 @@
         "/property:GenerateFullPaths=true",
         "/consoleloggerparameters:NoSummary"
       ],
-      "problemMatcher": "$msCompile"
+      "problemMatcher": []
     },
     {
       "label": "publish",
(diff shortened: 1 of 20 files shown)


.vscode/tasks.json
```diff
@@ -11,7 +11,7 @@
         "/property:GenerateFullPaths=true",
         "/consoleloggerparameters:NoSummary"
       ],
-      "problemMatcher": "$msCompile"
+      "problemMatcher": []
```
jmarolf, line 14: do you know what this change is about?

JoeRobich: I should have talk about this one. So OmniSharp will report diagnostics in the Problems tab. Running a build using a problems matcher will report build warnings and errors in the Problemss tab. This will duplicate warning and errors in the Problems tab amd the build reported problems won't go away until a new build is run. So, removing the problemMatcher will keep duplicate, zombie issues from entering the Problems tab.

Issue #671, opened 2020-05-11
dotnet-format creates empty files if non-existant file is given with charset


legionaryu:
I'm using dotnet-format with Unity3D and for some reason every time I try to format using the project files generated by Unity3D , the dotnet-format creates some empty files and it breaks the Unity3D c-sharp compilation and I have to manually remove the empty files to make it work again.

For now, I'm formatting folders instead of the .sln/.csproj files.

jmarolf:
@legionaryu what are the names of these empty files? Can you provide steps to reproduce the issue?

legionaryu:
@jmarolf Just letting you know that I'm trying to isolate the issue because of the original project has sensitive information, so I cannot share it.
As soon as I'm able to reproduce in another environment I'll upload it here.

jmarolf:
Thanks @legionaryu! I understand that it can sometime be hard to reduce these cases. If you could tell me the version of `dotnet format` you are using as well as the commandline args that would be a great start.

legionaryu:
Hi @jmarolf!
Good news, I was able to reproduce the issue with the attached project. 
[EmptyFile.zip](https://github.com/dotnet/format/files/4611123/EmptyFile.zip)

My version of dotnet-format is 3.3.111304+7c8f67a570f5fde6a247704733d6742f93c0fa48

From my tests, I was able to conclude that if a non-existant file is referenced in the csproj and the .editorconfig explicitly set a charset, the formatter creates an empty file with its name.

Here my csproj:
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp5.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="Empty.cs" /> <!--This file DOESN'T exists-->
  </ItemGroup>

</Project>
```

And here is my .editorconfig:
```
root = true

[*.*]
charset = utf-8
```

jmarolf:
@legionaryu thanks! I've updated the title to reflect this.

PR #775, opened 2020-08-26
Check that file exists before considering it for formatting

JoeRobich:
Resolves https://github.com/dotnet/format/issues/671

### DIFF
--- a/src/CodeFormatter.cs
+++ b/src/CodeFormatter.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
+using System.IO;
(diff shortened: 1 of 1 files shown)


src/CodeFormatter.cs
```diff
@@ -197,7 +198,7 @@ private static Workspace OpenFolderWorkspace(string workspacePath, SourceFileMat
                     addedFilePaths.Add(document.FilePath);
 
                     var isFileIncluded = formatOptions.WorkspaceType == WorkspaceType.Folder ||
```
JoeRobich, line 200: Documents in a folder workspace have already been verified to exist.

jmarolf (approved):

PR #786, opened 2020-08-28
Remove dangling argument from changelog example

iinuwa:
I think the `-f` is not needed/incorrect here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@
 
 #### Deprecations:
 - Added warning to use the default argument instead of `--workspace` option. Use `dotnet format ./format.sln` instead of `dotnet format -w ./format.sln`
-- Added warning to use the default argument to specify the folder path when using the `--folder` option. Use `dotnet format ./src -f` instead of `dotnet format -f ./src`
+- Added warning to use the default argument to specify the folder path when using the `--folder` option. Use `dotnet format ./src` instead of `dotnet format -f ./src`
 - Added warning to use `--include` instead of `--files` alias.
 - Added warning to use `--check` instead of `--dry-run` alias.
 


JoeRobich:
> Use `dotnet format ./src -f` instead of `dotnet format -f ./src`

@iinuwa,
In the 3.x builds of dotnet-format the correct commandline for working against a folder of files was `dotnet format -f ./src`. The folder path was an argument to the `-f` option.
in the 4.x builds of dotnet-format we transitioned to taking the path as an argument to the tool itself, so the correct commandline for working against a folder of files is `dotnet format ./src -f`. Leaving off the `-f` option in this case would cause the tool to look for a solution or project file to work against in the ./src folder.

am11:
Sorry for the intrusion.
> the correct commandline for working against a folder of files is `dotnet format ./src -f`

Few hours ago, I was working on dotnet/runtime repo and found that `-f` option (with `v4.1.131201+82a5527cd93d66bffc885c0bb901499f5981db35`) is working as described for some reason.

```sh
# bash on macOS
> git clone https://github.com/dotnet/runtime --single-branch --depth 1
> dotnet-format runtime/src/libraries/System.Drawing.Common/src/System/Drawing/ --folder
  Unable to locate MSBuild. Ensure the .NET SDK was installed with the official installer.

# return code
> $?
3
```

is this a known limitation? I copied the changed file (https://github.com/dotnet/runtime/pull/41503) to `/tmp` and ran formatter `dotnet-format /tmp --folder`, and then copied back as a workaround.

JoeRobich:
Up until a recent change in the 5.x branches, we would look for MSBuild even when using the --folder option.

am11:
`dotnet-format` only gave me error in dotnet/runtime's subdirectory, otherwise my installation is functional on this system (msbuild is present, it works for my own projects and `/tmp` example above). Perhaps it conflicts with SDK version in dotnet/runtime's global.json (even when calling from outside of source tree, as shown above). In which case, the error message is bit misleading. Would be nice if it can support such scenarios or provide accurate error message.

JoeRobich:
@am11 Please open an issue for this and we can investigate.

@iinuwa I am going to close this PR since the documentation is correct for formatting a folder of files. Thanks!

iinuwa:
> > Use `dotnet format ./src -f` instead of `dotnet format -f ./src`
> 
> @iinuwa,
> In the 3.x builds of dotnet-format the correct commandline for working against a folder of files was `dotnet format -f ./src`. The folder path was an argument to the `-f` option.
> in the 4.x builds of dotnet-format we transitioned to taking the path as an argument to the tool itself, so the correct commandline for working against a folder of files is `dotnet format ./src -f`. Leaving off the `-f` option in this case would cause the tool to look for a solution or project file to work against in the ./src folder.
> 

Got it. Thanks!

PR #790, opened 2020-08-30
Add ability to read --{in,ex}clude value from stdin

am11:
Today, `dotnet-format` accepts space-separated list of files. This
allows us to pass files via shell's native globbing expansion,
heredoc style redirections as well as via pipelines (see #551 for
examples). However, in case of pipeline it requires a second utility
(`xarg`) to transform pipeline input as space-separated list to
`dotnet-format`.

This PR implements native pipeline reading support for `--include`
and `--exclude` options, while keeping the space-separated list
intact.

#### Usage examples

1. Include all C# source files under `tests/Utilities` directory
that are in git source tree:

    ```sh
    git ls-files :/tests/Utilities/*.cs | dotnet format --include - --folder
    ```
2. Format all C# source files in last commit:

    ```sh
    git show --name-only --pretty="" | dotnet format --include - --folder
    ```
2. Exclude certain `*.cs` and `*.vb` files using `ls(1)`:

    ```sh
    ls ../../../../../src/generated/{*.cs,*.vb} | dotnet format --exclude /dev/stdin --folder
    ```

#### Rules

* Based on Guideline 13 of [IEEE 1003.1-2017](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html#tag_12_02) (POSIX),
  it accepts `-` as an explicit marker for stdin with
addition of:
  * `/dev/stdin` - which is a universal synonym of `-` on all Unices.
  * It *only* accepts explicit markers (`/dev/stdin` or `-`) and
  does not implicitly deduce the output if standard input was redirected,
  but marker was not present. This is because our usage is multi-purpose
  (both `--include` and `--exclude`).
* It is an error if both `--include` and `--exclude` are using stdin
marker (`/dev/stdin` or `-`).

#### Limitations / future considerations

* Currently, it reads the entire input from pipeline in `include`/`exclude`
buffer, and then runs the operation on the whole batch. In order
to make it true pipeline friendly, it would require some refactoring;
so files are `yield return`'d and enumerator can dispatch format
operation per file.
* At present, we do not have out-of-process functional test mechanism
for CLI to effectively validate these kind of use-cases (redirection
and shell globbing vs. dotnet-format's built-in globbing support);
so no tests are included in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,19 +85,25 @@ Options:
 
 Add `format` after `dotnet` and before the command arguments that you want to run:
 
-| Examples                                                         | Description                                                                                   |
-| ---------------------------------------------------------------- |---------------------------------------------------------------------------------------------- |
-| dotnet **format**                                                | Formats the project or solution in the current directory.                                     |
-| dotnet **format** &lt;workspace&gt;                              | Formats a specific project or solution.                                                       |
-| dotnet **format** &lt;workspace&gt; -f                           | Formats a particular folder and subfolders.                                                   |
-| dotnet **format** &lt;workspace&gt; --fix-style warn             | Fixes only codestyle analyzer warnings.                                                            |
-| dotnet **format** &lt;workspace&gt; --fix-whitespace --fix-style | Formats and fixes codestyle analyzer errors.                                                  |
-| dotnet **format** &lt;workspace&gt; --fix-analyzers              | Fixes only 3rd party analyzer errors.                                                              |
-| dotnet **format** &lt;workspace&gt; -wsa                         | Formats, fixes codestyle errors, and fixes 3rd party analyzer errors.                         |
-| dotnet **format** -v diag                                        | Formats with very verbose logging.                                                            |
-| dotnet **format** --include Programs.cs Utility\Logging.cs       | Formats the files Program.cs and Utility\Logging.cs                                           |
-| dotnet **format** --check                                        | Formats but does not save. Returns a non-zero exit code if any files would have been changed. |
-| dotnet **format** --report &lt;report-path&gt;                   | Formats and saves a json report file to the given directory.                                  |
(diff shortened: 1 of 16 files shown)


jmarolf:
@jonsequitur how difficult would it be to get support for this built into System.Commandline

jonsequitur:
We've been wanting to add support for stdin. This is something we've wanted to do. I don't think it would be too difficult, but we'd want to talk through the design goals a bit to come up with something generalizable and that works with the model binding infrastructure.

jmarolf:
@jonsequitur got it. My preference would be to have reading from std in be a feature of system.commandline rather than us needing to special case it. I am ok with this PR going in with the assumption that we will remove this code once a more generalized solution is available.

jonsequitur:
Agreed. 

Related, this issue has been open a while: https://github.com/dotnet/command-line-api/issues/275.

am11:
Resolved merge conflicts by other PR merge.

> a more generalized solution is available.

That would be great, indeed. I think stdin use cases are typically a bit tricky (but likely not impossible) to generalize, as it entails number of customizations:

* is stdin enabled for multiple options
* which of the stdin-accepting options are they mutually exclusive
* is consumer streaming in data (which was the case for roslyn: https://github.com/dotnet/roslyn/pull/41166) or blocked-reading (this PR)

etc.

am11:
@jmarolf, I will keep an eye on stdin support upstream and adapt when it happens. If these changes are fine otherwise, could this PR be merged? Thanks!

jmarolf (approved):

jmarolf:
@am11 thanks for pinging me no this. @JoeRobich can you review this as well? If you signoff I'll merge this.

JoeRobich (approved):

PR #864, opened 2020-11-16
When matching all files don't rely on FileMatcher

### LINKED ISSUE #809: Are roslyn analyzers working?
I'm not sure if Roslyn analyzers should already work in current version 5.0.146101. I have tried those:

- from docs (https://github.com/dotnet/format/blob/master/docs/Supported-.editorconfig-options.md)

dotnet_diagnostic.IDE0005.severity = warning

- and one from (https://github.com/DotNetAnalyzers/StyleCopAnalyzers) for blank lines. SA1507: Code should not contain multiple blank lines in a row

dotnet_diagnostic.SA1507.severity = warning

Visual studio is showing hints for fixing errors but after run:

dotnet format --fix-style warn

nothing is fixed ;(

JoeRobich:
When Including "**/\*.\*" (All files in all directories), the file matcher doesn't return matches for valid filenames. This PR skips checking the file matcher when we are matching all files.

Resolves #809

### DIFF
--- a/src/CodeFormatter.cs
+++ b/src/CodeFormatter.cs
@@ -194,7 +194,7 @@ namespace Microsoft.CodeAnalysis.Tools
                     addedFilePaths.Add(document.FilePath);
 
                     var isFileIncluded = formatOptions.WorkspaceType == WorkspaceType.Folder ||
-                        (formatOptions.FileMatcher.Match(document.FilePath).HasMatches && File.Exists(document.FilePath));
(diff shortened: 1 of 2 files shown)


jmarolf (approved):

JoeRobich:
Opened https://github.com/dotnet/runtime/issues/44767 to track this bug with FileSystemGlobbing.Matcher.

Issue #67, opened 2019-03-26
indentation and blank line


OneCyrus:
please add consistent cleanup for blank lines and indentation. it would help a lot to enforce consistent code style.

jmarolf:
@OneCyrus do you want new editorconfig rules to enforce this?  What would the design look like?

OneCyrus:
for indentation it should respect the editorconfig tab/spaces rule. additionally just a flag to enable/disable it. i guess for identation there aren't too many right ways. just stick to the way it's done in the microsoft c# coding style guidelines.

for blank lines there should be two rules:
- no consecutive blank lines ([see tslint](https://palantir.github.io/tslint/rules/no-consecutive-blank-lines/))
- blank line before code block ([see tslint plugin](https://github.com/xbdtb/tslint-plugin-blank-line))

AnthonySteele:
The rules for c# code that I recall from stylecop are simple and neaten code considerably. i would like to see them back as a default or failing that, as an option. They are:

* Two or more blank lines in a row: Remove down to one blank line.
* Blank line immediately after a `{` line or before a `}` line: remove it.

OneCyrus:
it looks like most StyleAnalyzer rules have code fixes. How about leverage them and let the user customize the rules (shipped with an opinionated default config)?

https://dotnetanalyzers.github.io/StyleCopAnalyzers/

JoeRobich:
Analyzer support is being tracked by https://github.com/dotnet/format/issues/416

ClankBot:
Hey, nice work on this tools, however after a few uses i noticed there is no blank line removal

Any update on ?

> for blank lines there should be two rules:
> - no consecutive blank lines (see tslint)
> - blank line before code block (see tslint plugin)

bigdnf:
I also want to know is it possible in current version to remove blank lines?

dzmitry-lahoda:
@VladimirGlushets we still do not have accepted new line formatting.

JoeRobich:
Analyzer support has been implemented and StyleCopAnalyzers work for removing empty lines. See this comment https://github.com/dotnet/format/issues/1074#issuecomment-814260542

PR #884, opened 2020-12-02
Improve FixAll support by using equivalence key when available

JoeRobich:
While investigating the fixes for these blank line StyleCop rules (https://github.com/dotnet/format/issues/67#issuecomment-478958324), I saw format was reporting no fixall action returned from the provider. Investigating the StyleCop repo it seemed that an action should be returned. After verifying that the VS Code C# extension could apply a solution fixall for those diagnostics, I looked for differences in implementation. It seems necessary to provide the equivalence key for some fixall providers.

### DIFF
--- a/src/Analyzers/SolutionCodeFixApplier.cs
+++ b/src/Analyzers/SolutionCodeFixApplier.cs
@@ -29,37 +29,57 @@ namespace Microsoft.CodeAnalysis.Tools.Analyzers
                 return solution;
             }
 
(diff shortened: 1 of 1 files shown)


jmarolf (approved):

jmarolf:
@JoeRobich do we want to try and use the `IDiagnosticAnalyzerService` so we don't need to re-implement the set of things codefixes need?

Issue #717, opened 2020-07-02
Add unit tests for AnalyzerFormatter


jmarolf:

JoeRobich:
A few cases that I can think of:
- How we check an analyzer's potential severity to determine whether it should be included in analysis
  - When it is configured via a codestyle .editorconfig option name
  - When it is configured via a 'dotnet_diagnostic.{id}.severity' option
  - When it is configured via a 'dotnet_analyzer_diagnostic.category-{name}.severity' category option
  - When it is configured via the 'dotnet_analyzer_diagnostic.severity' bulk option
- How we apply code fixes to ensure
  - We warn when a Solution scope fixall is not available
  - We warn when a fixall throws an exception
  - We run fixers for CodeStyle diagnostics
  - We run fixers for 3rd part diagnostics
  - We run fixers for Compiler diagnostics
  - We run all fixalls for a diganostic id (in the case that C# and VB share a diagnostic id and have separate fixers)

PR #896, opened 2020-12-11
Add unit test for 3rd party fixer formatting

JoeRobich:
Adds infrastructure for testing that 3rd party fixes are applied during formatting.

Part of the work for #717

### DIFF
--- a/src/Workspaces/MSBuildWorkspaceLoader.cs
+++ b/src/Workspaces/MSBuildWorkspaceLoader.cs
@@ -13,6 +13,8 @@ namespace Microsoft.CodeAnalysis.Tools.Workspaces
 {
     internal static class MSBuildWorkspaceLoader
     {
+        private static readonly SemaphoreSlim s_guard = new SemaphoreSlim(1, 1);
+
         public static async Task<Workspace?> LoadAsync(
             string solutionOrProjectPath,
             WorkspaceType workspaceType,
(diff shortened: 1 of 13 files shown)


src/Workspaces/MSBuildWorkspaceLoader.cs
```diff
@@ -29,35 +31,45 @@ internal static class MSBuildWorkspaceLoader
                 { "AlwaysCompileMarkupFilesInSeparateDomain", bool.FalseString },
             };
 
-            var workspace = MSBuildWorkspace.Create(properties);
+            MSBuildWorkspace workspace;
 
-            Build.Framework.ILogger? binlog = null;
-            if (createBinaryLog)
+            await s_guard.WaitAsync();
```
JoeRobich, line 36: This is needed because tests run in parallel but msbuild won't

tests/Analyzers/CodeStyleAnalyzerFormatterTests.cs
```diff
@@ -6,13 +6,19 @@
...
 
 namespace Microsoft.CodeAnalysis.Tools.Tests.Analyzers
 {
     public class CodeStyleAnalyzerFormatterTests : CSharpFormatterTests
     {
         private protected override ICodeFormatter Formatter => AnalyzerFormatter.CodeStyleFormatter;
 
+        public CodeStyleAnalyzerFormatterTests(ITestOutputHelper output)
+        {
+            TestOutputHelper = output;
```
JoeRobich, line 19: On failure of any tests in the class, the formatter log will be written to the test output.

src/Workspaces/MSBuildWorkspaceLoader.cs
jmarolf: Still don't quite understand this. Why can't each test create their own workspace?

jmarolf (approved):

src/Workspaces/MSBuildWorkspaceLoader.cs
JoeRobich: Since the CodeFormatterTests is creating MSBuildWorkspaces and now we are creating a MSBuildWorkspace here, there is an InvalidOperationException "Cannot start a new build while a build is in progress"

jmarolf: Could we create a `WorspaceFact` that controls this intead of changing product code?

JoeRobich: Took a stab at this in https://github.com/dotnet/format/pull/899

PR #945, opened 2021-01-29
Search .editorconfig only in relevant directories

am11:
Currently we are searching .editorconfig in:

1. child directories
2. parent directories up to the root

However, roslyn only performs <span>#</span>2: [ProjectExtensions.cs#L51](https://github.com/dotnet/roslyn/blob/d66c12164c0da3841f86b165872ea1b55942c575/src/Workspaces/SharedUtilitiesAndExtensions/Workspace/Core/Extensions/ProjectExtensions.cs#L51). Editor config's official docs also attest to that being the only criteria as far as far as "walking the filesystem" goes: https://editorconfig.org/#file-location.

The typical configuration systems in many tools of linter, compiler and transpiler categories use the following mechanism with intuitive/easy-to-remember overriding/precedence rules for developers:

1. accept config file path via command-line argument and use that.
2. if <span>#</span>1 is not present, then look for the environment variable, e.g. DOTNET_EDITORCONFIG_PATH or simply EDITORCONFIG_PATH.
3. if <span>#</span>2 is not present, then look in parent directories up to the volume root.
4. if <span>#</span>3 is not present, then look in $HOME directory.

As roslyn also has a TODO to improve the heuristics and currently only performs search in parent directories per the docs, this PR aims to only fix the bug. In future, we may want to improve the heuristics based on the aforementioned well-known convention.

Before it was taking 22425ms to format _one_ .cs file in a directory with nested subdirectories on my system. After this fix, it took 1351ms for the same file at the same location.

### DIFF
--- a/src/Utilities/EditorConfigFinder.cs
+++ b/src/Utilities/EditorConfigFinder.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
 
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.IO;
 
@@ -7,39 +8,34 @@ namespace Microsoft.CodeAnalysis.Tools.Utilities
 {
     internal static class EditorConfigFinder
     {
-        public static ImmutableArray<string> GetEditorConfigPaths(string path)
+        public static ImmutableArray<string> GetEditorConfigPaths(ImmutableArray<string> paths)
         {
-            // If the path is to a file then remove the file name and process the
-            // folder path.
-            var startPath = Directory.Exists(path)
-                ? path
-                : Path.GetDirectoryName(path);
-
-            if (!Directory.Exists(startPath))
+            if (paths == default)
             {
                 return ImmutableArray<string>.Empty;
             }
 
             var editorConfigPaths = ImmutableArray.CreateBuilder<string>(16);
+            var visited = new List<string>();
 
-            var directory = new DirectoryInfo(path);
-
-            // Find .editorconfig files contained unders the folder path.
-            var files = directory.GetFiles(".editorconfig", SearchOption.AllDirectories);
-            for (var index = 0; index < files.Length; index++)
+            foreach (var path in paths)
             {
-                editorConfigPaths.Add(files[index].FullName);
-            }
+                var directoryName = Path.GetDirectoryName(path);
+                if (visited.Contains(directoryName)) continue;
 
-            // Walk from the folder path up to the drive root addings .editorconfig files.
-            while (directory.Parent != null)
-            {
-                directory = directory.Parent;
+                var directory = new DirectoryInfo(directoryName);
 
-                files = directory.GetFiles(".editorconfig", SearchOption.TopDirectoryOnly);
-                if (files.Length == 1)
+                // Walk from the folder path up to the drive root addings .editorconfig files.
+                while (directory.Parent != null)
                 {
-                    editorConfigPaths.Add(files[0].FullName);
+                    if (visited.Contains(directory.FullName)) break;
+
+                    visited.Add(directory.FullName);
+
+                    var file = Path.Combine(directory.FullName, ".editorconfig");
+                    if (File.Exists(file)) editorConfigPaths.Add(file);
+
+                    directory = directory.Parent;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


JoeRobich:
@am11 Sorry this is causing performance issues for you and thanks for taking a look into it. For this to truly be correct. We should determine the set of paths containing formattable files. Then, we should walk up from each of those paths. If we only walk up from the workspace folder path, then we will miss .editorconfigs that exist above code files but beneath workspace folder.

xt0rted:
@JoeRobich does dotnet format handle [Global AnalyzerConfigs](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#global-analyzerconfig)? I haven't tried them with it yet, but the linked targets file looks related to how I'm adding them in with my `Directory.Build.targets`:

```xml
<Project>

  <Target Name="AddGlobalCodeAnalysisRules" BeforeTargets="CoreCompile">
    <ItemGroup>
      <EditorConfigFiles Include="$(MSBuildThisFileDirectory)analysis\*.editorconfig" />
    </ItemGroup>
  </Target>

</Project>
```

JoeRobich:
@xt0rted We do when not using the `--folder` flag. The EditorConfigFinder would need to be updated to similarly look for '.globalconfig' files. Replicating this bit of MSBuild logic.

https://github.com/dotnet/roslyn/blob/8637000fc35f01a453349147b71d5ecdae8d0bff/src/Compilers/Core/MSBuildTask/Microsoft.Managed.Core.targets#L88-L89

am11:
@JoeRobich, thanks for the reminder. I was using `--folder` option with single workspace (with accompanying `--include` option).

Please correct me if I am wrong, but I think it is a bit more involved. e.g. given:

```
{root}
    \
    .editorconfig
    dir1
        \ 
        file1.cs
        .editorconfig
        dir2
            \ 
            file2.cs
            dir3
                \ 
                file3.cs
                .editorconfig
```

To format `file3.cs`, currently we are applying `{root}\dir1\dir2\dir3\.editorconfig`, `{root}\dir1\.editorconfig` and `{root}\.editorconfig`. It seems like we should only be applying: `{root}\dir1\dir2\dir3\.editorconfig`.

Similarly, for `file1.cs` and `file2.cs`, we should only applying `{root}\dir1\.editorconfig`, rather than `{root}\dir1\.editorconfig` _and_ `{root}\.editorconfig`.

To support that, we would need to implement file:config grouping for each source file and optimize algorithm to avoid walking same fragment of path multiple times.

JoeRobich:
@am11 The compiler takes this list of potential .editorconfigs and then determines which to apply to a particular code file based on parent folders and whether the .editorconfig applies the `root = true` option. If none of the .editorconfigs in your example contained the `root = true` then the options from all .editorconfigs would be applied to file3.cs.

am11:
> We should determine the set of paths containing formattable files. Then, we should walk up from each of those paths.

@JoeRobich, I have implemented the strategy with a uniqueness constraint that a segment of the path will be visited once.

src/Utilities/EditorConfigFinder.cs
```diff
@@ -1,45 +1,41 @@
...
 
-                files = directory.GetFiles(".editorconfig", SearchOption.TopDirectoryOnly);
-                if (files.Length == 1)
+                // Walk from the folder path up to the drive root addings .editorconfig files.
+                while (directory.Parent != null)
                 {
-                    editorConfigPaths.Add(files[0].FullName);
+                    directory = directory.Parent;
+
+                    var files = directory.GetFiles(".editorconfig", SearchOption.TopDirectoryOnly);
```
JoeRobich, line 34: We should check whether the directory has been visited and add the directory to visited within this loop, since many source directories will share the same branches.

src/Utilities/EditorConfigFinder.cs
```diff
@@ -1,45 +1,41 @@
...
+                visited.Add(directoryName);
+                var directory = new DirectoryInfo(directoryName);
 
-                files = directory.GetFiles(".editorconfig", SearchOption.TopDirectoryOnly);
-                if (files.Length == 1)
+                // Walk from the folder path up to the drive root addings .editorconfig files.
+                while (directory.Parent != null)
                 {
-                    editorConfigPaths.Add(files[0].FullName);
+                    directory = directory.Parent;
```
JoeRobich, line 32: Immediately going to the parent directory will cause us to skip .editorconfig files that live beside code files.

JoeRobich (commented):
A couple issues to resolve then we can take this change. Thanks!

am11:
@JoeRobich, unfortunately the new approach has significantly regressed the performance for `dotnet format --folder`, where current directory has large number of sub(-sub*)directories (which was my use-case; to run that command from repo root with `--include` filter argument, e.g. files being changed in the PR).

Basically the issue boils down to this line:
https://github.com/dotnet/format/blob/dbec967f9929e3cf257cb02d0ebcb79a3c1c529b/src/Utilities/EditorConfigFinder.cs#L28
if we replace it with manual loop or add filters (using EnumerateFiles etc.), to implement `keep finding in child directories as is; BUT only where it intersects with the given set of file paths`, implementing that `BUT` part regresses in performance and there seems to be no good performance-friendly API to support it.

So the current implementation in master branch is most likely the best what we can do right now. Thoughts?

Issue #971, opened 2021-02-12
`dotnet format --fix-style` reports errors that `dotnet build` does not


bradwilson:
1. Clone https://github.com/xunit/xunit
2. Edit `.config/dotnet-tools.json` and update `dotnet-format` to `5.0.211103`
3. Run `dotnet tool restore`

Run a normal build: `dotnet build`. Note there are no build warnings or errors:

```
[...]
  xunit.v3.runner.tdnet.tests -> C:\Dev\xunit\xunit\src\xunit.v3.runner.tdnet.tests\bin\Debug\net472\xunit.v3.runner.tdnet.tests.exe
  xunit.v3.common.tests -> C:\Dev\xunit\xunit\src\xunit.v3.common.tests\bin\Debug\netcoreapp2.1\xunit.v3.common.tests.dll
  xunit.v3.runner.inproc.console.tests -> C:\Dev\xunit\xunit\src\xunit.v3.runner.inproc.console.tests\bin\Debug\netcoreapp2.1\xunit.v3.runner.inproc.console.tests.dll
  xunit.v3.runner.utility.tests -> C:\Dev\xunit\xunit\src\xunit.v3.runner.utility.tests\bin\Debug\netcoreapp2.1\xunit.v3.runner.utility.tests.dll
  xunit.v3.assert.tests -> C:\Dev\xunit\xunit\src\xunit.v3.assert.tests\bin\Debug\netcoreapp2.1\xunit.v3.assert.tests.dll
  xunit.v3.runner.msbuild.tests -> C:\Dev\xunit\xunit\src\xunit.v3.runner.msbuild.tests\bin\Debug\net472\xunit.v3.runner.msbuild.tests.exe

Build succeeded.
    0 Warning(s)
    0 Error(s)
```

Now run `dotnet format --fix-style`. It appears to be reporting nullability errors that a normal build does not:

```
  Formatting code files in workspace 'C:\Dev\xunit\xunit\xunit.sln'.
  Warnings were encountered while loading the workspace. Set the verbosity option to the 'diagnostic' level to log warnings.
  src\xunit.v3.assert\Asserts\Sdk\AssertEqualityComparerAdapter.cs(38,32): Converting null literal or possible null value to non-nullable type. (CS8600)
  src\xunit.v3.assert\Asserts\Sdk\AssertEqualityComparerAdapter.cs(38,38): Converting null literal or possible null value to non-nullable type. (CS8600)
[...]

PR #981, opened 2021-02-16
Only report fixable compiler diagnostics.

JoeRobich:
Partially addresses the issues raised in https://github.com/dotnet/format/issues/971.

This will filter out compiler diagnostics that do not have an associated CodeFixProvider.


jmarolf (approved):
:shipit:

PR #1007, opened 2021-02-26
Add option to filter diagnostics by id

### LINKED ISSUE #1006: Ability to filter diagnostics being fixed by specifying a list of diagnostic ids
Add a `--diagnostics` option which takes a space separated list of diagnostic ids. This list of ids will be used to filter out which code style or third-party analyzers are run. This would work in conjunction with the specified fix severity.

Would help in situations like https://stackoverflow.com/questions/66307987/how-to-script-removal-of-redundant-using-directives-from-the-given-c-sharp-sourc/66322867?noredirect=1#comment117320368_66322867 where the user wishes to fix a single diagnostic, in this case IDE0005.

JoeRobich:
Resolves #1006

### DIFF
--- a/src/FormatCommand.cs
+++ b/src/FormatCommand.cs
@@ -89,7 +89,7 @@ namespace Microsoft.CodeAnalysis.Tools
             var folder = symbolResult.ValueForOption<bool>("--folder");
             var fixAnalyzers = symbolResult.OptionResult("--fix-analyzers");
             return folder && fixAnalyzers != null
-                ? "Cannot specify the '--folder' option when running analyzers."
+                ? Resources.Cannot_specify_the_folder_option_when_running_analyzers
                 : null;
         }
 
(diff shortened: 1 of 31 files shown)


src/FormatCommand.cs
```diff
@@ -49,6 +50,10 @@ internal static RootCommand CreateCommandLineOptions()
                 {
                     Argument = new Argument<string?>("severity") { Arity = ArgumentArity.ZeroOrOne }.FromAmong(SeverityLevels)
                 },
+                new Option(new[] { "--diagnostics" }, Resources.A_space_separated_list_of_diagnostic_ids_to_use_as_a_filter_when_fixing_code_style_or_3rd_party_issues)
+                {
+                    Argument = new Argument<string[]>(() => Array.Empty<string>())
```
jmarolf, line 55: I would like it if we could provide a helpful message if this option is provide and we are not in a `--fix-stlye` or `--fix-analyzers` mode

jmarolf (approved):

PR #1021, opened 2021-02-28
Reenabled .NET Core 2.1 support

### LINKED ISSUE #986: Latest format tool v5.0.211103 does not work with .NET Core 2.1.25
Hi @JoeRobich

When I installed the latest format tool v5.0.211103, and my machine only has .NET Core 2.1.25 installed, `dotnet format` does not work.

It looks like caused by the following change, when I removed `-rc1` from the config file, it works fine.
[https://github.com/dotnet/format/blob/eadde172822e75922a68332d1037972f1fcec791/src/dotnet-format.csproj#L19](https://github.com/dotnet/format/blob/eadde172822e75922a68332d1037972f1fcec791/src/dotnet-format.csproj#L19)

issue

JoeRobich:
Resolves #986

In its current state dotnet-format will run on only pre-release .NET Core 2.1 and release or pre-release .NET Core 3.1 or higher. Removing the `-rc1` from the runtime version will allow it to run on release versions of 2.1 again but removes the ability to run against pre-release versions.

We should take this change for 5.1. Make 5.1 a LTS release that will be supported with bug fixes while .NET Core 2.1 is under support. Then, we should bump the version number to 6.0 and move the framework version to a 3.1 pre-release dropping 2.1 support from the 6.0 release.


jmarolf (approved):

jmarolf (approved):

JoeRobich:
Failing integration tests are because this build cannot rollforward to preview SDKs. Unable to find a suitable commit for some repos. Removing them from the 5.x integration tests.

Issue #1041, opened 2021-03-10
The tool creates formatDiagnosticLog.binlog in the solution directory and no way to disable this or redirect


vitek-karas:
Running `dotnet-format some.sln --verbosity diagnostic` will create a file called `formatDiagnosticLog.binlog` next to the solution file. There doesn't seem to be a way to redirect this to another directory, rename it or disable it.

We want to keep diagnostic output to be able to detect issues with the solution loading, but adding files next to the solution makes the experience pretty bad for local enlistments since git will pick it up as a new file to be added.

Ideally there would be a way to disable writing of the file from command line, or at the very least ability to specify the file path to write it to.

JoeRobich:
@jmarolf I think it is time to add a `--binarylog <filename>` option.

PR #1044, opened 2021-03-11
Add separate command for binary log

jmarolf:
fixes https://github.com/dotnet/format/issues/1041

### DIFF
--- a/src/FormatCommand.cs
+++ b/src/FormatCommand.cs
@@ -3,7 +3,6 @@
 using System;
 using System.CommandLine;
 using System.CommandLine.Parsing;
-using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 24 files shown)


src/FormatCommand.cs
```diff
@@ -75,6 +77,11 @@ internal static RootCommand CreateCommandLineOptions()
                 {
                     IsHidden = true
                 },
+                new Option(new[] {"--binarylog", "-bl" }, Resources.Log_all_project_or_solution_load_information_to_a_binary_log_file)
+                {
+                    Argument = new Argument<string?>(
+                        () => Path.Combine(Environment.CurrentDirectory, "formatDiagnosticLog.binlog") ) { Name = "binary-log-path" }.LegalFilePathsOnly()
```
JoeRobich, line 83: This doesn't set it up to always provide a binarylog path right?

src/FormatCommand.cs
```diff
@@ -75,6 +77,11 @@ internal static RootCommand CreateCommandLineOptions()
                 {
                     IsHidden = true
                 },
+                new Option(new[] {"--binarylog", "-bl" }, Resources.Log_all_project_or_solution_load_information_to_a_binary_log_file)
```
JoeRobich, line 80: Although `-bl` is what msbuild uses, I think we want to stick to single letter short names.

JoeRobich: I wouldn't bother giving this one a short name
```suggestion
                new Option(new[] { "--binarylog" }, Resources.Log_all_project_or_solution_load_information_to_a_binary_log_file)
```

src/Program.cs
```diff
@@ -182,7 +183,7 @@ private static async Task<int> Main(string[] args)
                     formatOptions,
                     logger,
                     cancellationTokenSource.Token,
-                    createBinaryLog: logLevel == LogLevel.Trace).ConfigureAwait(false);
+                    binaryLogPath: binarylog).ConfigureAwait(false);
```
JoeRobich, line 186: If including the default value causes it to always provide a binarylog path, you could check for the option specifically.

```suggestion
                    binaryLogPath: s_parseResult.WasOptionUsed("--binarylog") ? binarylog : null).ConfigureAwait(false);
```

JoeRobich: It may be better to not provide a default value for the option and to set a default if `binarylog is null` and the `--binarylog` option was used...

src/FormatCommand.cs
jmarolf: my understanding is that this _should_ make it so `--binarylog` without a path just puts it in the current directory. I'll add a test for this

jmarolf: added tests and updated this

JoeRobich (approved):
Just need to update the readme and maybe the docs too, if you want to explain when/why to use it.

JoeRobich:
All CI has passed but GitHub has been having issues today and hasn't updated. Merging.

PR #1047, opened 2021-03-12
Update integration tests SHAs

JoeRobich:


JoeRobich:
After reviewing the updated integration test logs. 
- CHARSET formatter was replacing the entire file and the TextChanged message would include the entire text. Updated to not generate a TextChange message when the full file is replaced.
- The Microsoft.VisualStudio.Threading.Analyzer.VisualBasic depends on the CSharp assembly but they are in separate folders. Updated the AnalyzerLoader to resolve from all assemblies in an analyzer package. Also, added caching based on the full analyzer path. Roslyn loading the same analyzer 180+ times was not performing well.

JoeRobich:
@jmarolf after reviewing the issues logged for the Roslyn Compilers.sln, I wonder if MSBuildWorkspace is able to properly handle IVTs.

jmarolf:
@JoeRobich can you file a bug on dotnet/roslyn and assign it to me to investigate? I don't want to forget

JoeRobich:
> assign it to me to investigate?

@jmarolf Opened https://github.com/dotnet/roslyn/issues/51910

PR #1071, opened 2021-03-31
Run CodeStyle formatter before removing unnecessary imports

### LINKED ISSUE #1069: Must run dotnet-format twice to remove all unnecessary imports
I want to use 'var' everywhere in my code and remove all unnecessary imports. But I must run dotnet-format twice to get the desired effect.

.editorconfig

```
root = true

[*.cs]
csharp_style_var_elsewhere = true:warning
csharp_style_var_for_built_in_types = true:warning
csharp_style_var_when_type_is_apparent = true:warning

dotnet_diagnostic.IDE0007.severity = warning
dotnet_diagnostic.IDE0005.severity = warning
```

my code

```
using System;
using ConsoleApp.Builders;
using ConsoleApp.Sample;

namespace ConsoleApp
{
    class Program
    {
        static void Main(string[] args)
        {
            X variable= XBuilder.Build();
        }
    }
}
```

first run of `dotnet-format --fix-style warn` remove `using System` and change `X variable= XBuilder.Build()` to `var variable= XBuilder.Build()`

only second run of `dotnet-format --fix-style warn` remove `using ConsoleApp15.Sample` (type X is from ConsoleApp.Sample).

Is it possible to run `dotnet-format` once and get the desired effect?

JoeRobich:
Resolves #1069

As pointed out, converting declarations to use `var` when fixing code style issues potentially creates unused imports.

### DIFF
--- a/src/CodeFormatter.cs
+++ b/src/CodeFormatter.cs
@@ -26,8 +26,8 @@ namespace Microsoft.CodeAnalysis.Tools
             new CharsetFormatter(),
             new OrganizeImportsFormatter(),
             AnalyzerFormatter.CodeStyleFormatter,
-            new UnnecessaryImportsFormatter(),
-            AnalyzerFormatter.ThirdPartyFormatter);
+            AnalyzerFormatter.ThirdPartyFormatter,
+            new UnnecessaryImportsFormatter());
 
         public static async Task<WorkspaceFormatResult> FormatWorkspaceAsync(
             FormatOptions formatOptions,
--- a/src/CodeFormatter.cs
+++ b/src/CodeFormatter.cs
@@ -25,8 +25,8 @@ namespace Microsoft.CodeAnalysis.Tools
             new EndOfLineFormatter(),
             new CharsetFormatter(),
             new OrganizeImportsFormatter(),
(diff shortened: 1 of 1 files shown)


jmarolf (approved):
You read my mind, This was literally my thoughts. as long as we run this last should be correct (for most of the cases)

JoeRobich:
> as long as we run this last

Hmm. We can make this truly run last. Didn't think about how 3rd party analyzers might create this same situation.

src/CodeFormatter.cs
```diff
@@ -25,8 +25,8 @@ internal static class CodeFormatter
             new EndOfLineFormatter(),
             new CharsetFormatter(),
             new OrganizeImportsFormatter(),
-            new UnnecessaryImportsFormatter(),
             AnalyzerFormatter.CodeStyleFormatter,
+            new UnnecessaryImportsFormatter(),
             AnalyzerFormatter.ThirdPartyFormatter);
```
JoeRobich, line 30:
```suggestion
            AnalyzerFormatter.ThirdPartyFormatter,
            new UnnecessaryImportsFormatter());
```

PR #1123, opened 2021-05-04
Move to .NET Core 3.1

JoeRobich:
For dotnet format 6.0 we will target .NET Core 3.1 as our minimum runtime and the 5.x builds will be maintained for users who require .NET Core 2.1 support.


NuGet.config
```diff
@@ -2,6 +2,7 @@
 <configuration>
   <packageSources>
     <clear />
+    <add key="dotnet3.1" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet3.1/nuget/v3/index.json" />
```
JoeRobich, line 5: Added so we can target a 3.1 preview runtime.

src/dotnet-format.csproj
```diff
@@ -19,7 +19,7 @@
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
 
     <!-- Always run on the latest runtime installed. -->
-    <RuntimeFrameworkVersion>2.1.0-rc1</RuntimeFrameworkVersion>
+    <RuntimeFrameworkVersion>3.1.0-preview3.19553.2</RuntimeFrameworkVersion>
```
JoeRobich, line 22: Targeting a preview will allow the rollforward mechanism to rollforward to previews.

tests/Analyzers/AnalyzerAssemblyGenerator.cs
```diff
@@ -19,6 +19,25 @@ namespace Microsoft.CodeAnalysis.Tools.Tests.Analyzers
 {
     public static class AnalyzerAssemblyGenerator
     {
+        private static readonly string s_basePath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+
+        private static async Task<IEnumerable<MetadataReference>> GetReferencesAsync()
+        {
+            var references = new List<MetadataReference>()
+            {
+                MetadataReference.CreateFromFile(Path.Combine(s_basePath, Path.GetFileName(typeof(ImmutableArray).Assembly.Location))),
```
JoeRobich, line 28: Reference the version of these assemblies deployed with our tests.

tests/Analyzers/AnalyzerAssemblyGenerator.cs
```diff
@@ -19,6 +19,25 @@ namespace Microsoft.CodeAnalysis.Tools.Tests.Analyzers
...
+                MetadataReference.CreateFromFile(Path.Combine(s_basePath, Path.GetFileName(typeof(SharedAttribute).Assembly.Location))),
+                MetadataReference.CreateFromFile(Path.Combine(s_basePath, Path.GetFileName(typeof(CSharpCompilation).Assembly.Location))),
+                MetadataReference.CreateFromFile(Path.Combine(s_basePath, Path.GetFileName(typeof(DiagnosticAnalyzer).Assembly.Location))),
+                MetadataReference.CreateFromFile(Path.Combine(s_basePath, Path.GetFileName(typeof(CodeFixProvider).Assembly.Location))),
+            };
+
+            var netcoreMetadataReferences = await ReferenceAssemblies.NetCore.NetCoreApp31.ResolveAsync(LanguageNames.CSharp, CancellationToken.None);
```
JoeRobich, line 35: Using the NetStandard2.0 references causes a version conflict with the System.Runtime version.

NuGet.config
jmarolf: I don't think this is something we should worry about or support. I would only care about support 6.0 preview runtimes right? I assume we want to be out of the business of supporting preview runtimes?

src/dotnet-format.csproj
jmarolf: got it, so this is necessary for it to rollforward to 6.0 previews

jmarolf (approved):

Issue #1115, opened 2021-04-29
Include diagnostic id in report


xt0rted:
I'd like to surface the issues that are found in each code file along with their diagnostic id, that way they can be fixed on an issue by issue basis. Without having the diagnostic id there's no way to fix individual issues by looking at the report.

My end goal is to group the report data by file path and get a list of each diagnostic id in it. Then the user can decide which diagnostic ids to pass to `dotnet format --diagnostics <ids> --include <files>`.

PR #1133, opened 2021-05-05
Add DiagnosticId to the format report

JoeRobich:
Resolves https://github.com/dotnet/format/issues/1115


xt0rted:
@JoeRobich does this change add the diagnostic id to the console output, or just the json report?

xt0rted:
Just noticed this was published to nuget so I was able to test it out. Looks like it is in the console output and report. Thanks for making this change.

Issue #1061, opened 2021-03-25
--fix-style warn --check unable to find Nuget packages


MdyPrince:
I am running format check by running following command `dotnet-format ./MySolution.sln --fix-style warn --check` . This solution contains few c# and c++ projects. I am using .editorconfig provided with this repository(dotnet/format) filtered only for [.cs] with severity set to warn for all rule. I have added reference of few Nuget packages to my .csproj as follow. 

![image](https://user-images.githubusercontent.com/53044446/112421180-9b777a80-8d54-11eb-945d-544f4ed1de4b.png)

**After Running the above command to check style format, getting following warnings.**

>  ContextInfo.cs(22,14): The type or namespace name 'JsonPropertyName' could not be found (are you missing a using directive or an assembly reference?) (CS0246)
 ContextInfo.cs(25,14): The type or namespace name 'JsonPropertyNameAttribute' could not be found (are you missing a using directive or an assembly reference?) (CS0246)
  ContextInfo.cs(25,14): The type or namespace name 'JsonPropertyName' could not be found (are you missing a using directive or an assembly reference?) (CS0246)
  JsonWriter.cs(12,24): The name 'JsonSerializer' does not exist in the current context (CS0103)
  NLogLoggerUtility.cs(10,57): The name 'NLog' does not exist in the current context (CS0103)
  FileCompressorUtility.cs(27,42): The name 'ZipFile' does not exist in the current context (CS0103)
  FileCompressorUtility.cs(27,64): The name 'ZipArchiveMode' does not exist in the current context (CS0103)
  NLogLoggerUtility.cs(28,34): The type or namespace name 'NLog' could not be found (are you missing a using directive or an assembly reference?) (CS0246)
  DotNetUtilities\LoggerUtility\NLogLoggerUtility.cs(29,35): The type or namespace name 'NLog' could not be found (are you missing a using directive or an assembly reference?) (CS0246)
  NLogLoggerUtility.cs(39,40): The name 'NLog' does not exist in the current context (CS0103)
  NLogLoggerUtility.cs(54,17): The name 'NLog' does not exist in the current context (CS0103)
  NLogLoggerUtility.cs(59,17): The name 'NLog' does not exist in the current context (CS0103)
  NLogLoggerUtility.cs(64,17): The name 'NLog' does not exist in the current context (CS0103)
  NLogLoggerUtility.cs(65,17): The name 'NLog' does not exist in the current context (CS0103)

[...]

PR #1367, opened 2021-10-08
Limit reported compiler diagnostics to those that support FixAllScope.Solution

JoeRobich:
There are times when dotnet-format reports compiler diagnostics when running Analyzers (See https://github.com/dotnet/format/issues/1061). This can be caused by missing unrestored dependencies but also by dotnet-format being unable to properly load the workspace. We had already limited the compiler diagnostics that we report to those that also have an related CodeFixProvider. This PR takes it a step further and filters out those that do not support FixAllScope.Solution.


JoeRobich:
@jmarolf please take a look

jmarolf (approved):

Issue #1386, opened 2021-10-15
`dotnet format` without arguments - show usage instead of stack trace


nil4:
Today I learned that `dotnet format` is included with .NET 6.0 SDK, so I ran `dotnet format`, hoping to find out which arguments it expects:

```cmd
> dotnet --version
6.0.100-rc.2.21505.57

> dotnet format
```

#### Observed results

Red stack trace text filled the console window, scrolling the point it was trying to make with the first line of output off the screen:  

```cmd
> dotnet format
Unhandled exception: System.IO.FileNotFoundException: Could not find a MSBuild project file or solution file in 'C:\some\path'. Specify which to use with the <workspace> argument.
   at Microsoft.CodeAnalysis.Tools.Workspaces.MSBuildWorkspaceFinder.FindWorkspace(String searchDirectory, String workspacePath)
   at Microsoft.CodeAnalysis.Tools.Workspaces.MSBuildWorkspaceFinder.FindWorkspace(String searchDirectory, String workspacePath)
   at Microsoft.CodeAnalysis.Tools.FormatCommandCommon.ParseWorkspaceOptions(ParseResult parseResult, FormatOptions formatOptions)
   at Microsoft.CodeAnalysis.Tools.Commands.RootFormatCommand.FormatCommandDefaultHandler.InvokeAsync(InvocationContext context)
   at System.CommandLine.Invocation.InvocationPipeline.<>c__DisplayClass4_0.<<BuildInvocationChain>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass23_0.<<UseParseErrorReporting>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass16_0.<<UseHelp>b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass27_0.<<UseVersionOption>b__1>d.MoveNext()
--- End of stack trace from previous location ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass25_0.<<UseTypoCorrections>b__0>d.MoveNext()
--- End of stack trace from previous location ---
[...]

PR #1387, opened 2021-10-15
Add try/catch around FindWorkspace calls to handle FileNotFound

JoeRobich:
Resolves https://github.com/dotnet/format/issues/1386


src/Commands/FormatAnalyzersCommand.cs
```diff
@@ -34,7 +35,16 @@ public async Task<int> InvokeAsync(InvocationContext context)
...
                 formatOptions = parseResult.ParseCommonOptions(formatOptions, logger);
-                formatOptions = parseResult.ParseWorkspaceOptions(formatOptions);
+
+                try
+                {
+                    formatOptions = parseResult.ParseWorkspaceOptions(formatOptions);
+                }
+                catch (FileNotFoundException fex)
+                {
+                    logger.LogError(fex.Message);
```
jmarolf, line 45: shouldn't we show a message saying we couldn't find the file instead of logging an error? Doesn't this still show a stack trace to the user?

jmarolf: since this is in `main` I am ok with waiting for a full re-work. do we want this to go into the 6.x branch instead?

jmarolf: alright, I'll email .NET Tactics to see if they want to take this.

PR #1396, opened 2021-10-20
Move from Features to CodeStyle libraries

JoeRobich:
a different approach

### DIFF
--- a/src/Analyzers/Extensions.cs
+++ b/src/Analyzers/Extensions.cs
@@ -9,21 +9,24 @@ using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis.Diagnostics;
-using Microsoft.CodeAnalysis.Options;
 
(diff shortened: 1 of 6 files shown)


src/Analyzers/Extensions.cs
```diff
@@ -152,11 +154,17 @@ public static DiagnosticSeverity ToSeverity(this ReportDiagnostic reportDiagnost
...
                 DiagnosticDescriptor descriptor,
                 Compilation compilation,
-                OptionSet options,
+                AnalyzerConfigOptions options,
                 out DiagnosticSeverity severity)
             {
                 severity = DiagnosticSeverity.Hidden;
 
+                if (options.GetType().GetField("_backing", BindingFlags.Instance | BindingFlags.NonPublic)
+                    ?.GetValue(options) is not IDictionary backingOptions)
```
jmarolf, line 163: genius

jmarolf: image

Issue #1419, opened 2021-11-04
dotnet-format should not pin to NET 6 sdks.


JoeRobich:
dotnet-format should run on the highest SDK that is available. Currently dotnet-format filters available MSBuild installations down to NET 6 SDKs.

PR #1420, opened 2021-11-04
Allow dotnet-format to run with future NET 7 SDKs

JoeRobich:
Resolves https://github.com/dotnet/format/issues/1419


jmarolf (approved):

src/Commands/FormatCommandCommon.cs
```diff
@@ -345,9 +345,9 @@ internal static bool TryLoadMSBuild([NotNullWhen(returnValue: true)] out string?
         {
             try
             {
-                // Get the latest .NET 6 SDK instance.
+                // Get the latest .NET SDK instance.
```
RikkiGibson, line 348: do we take this code path when running as a global tool? it seems like we would want "whatever SDK is launching us" in that case, right?

JoeRobich: We always take this path. The CLI and included tooling is typically updated to be the highest versioned SDK.

PR #1426, opened 2021-11-08
Fix small typo

monoman:


JoeRobich (approved):
Thanks for the fix! You found one of the alternate names we were considering as part of the move into the .NET SDK.

PR #1456, opened 2021-12-09
Include SourceGeneratedDocuments when including generated files

JoeRobich:
Source generated documents are not included in Project.Documents and have to be queried separately. Although it isn't typically useful to report issues for these documents, it can be helpful if they are generating part of your public API which is being tracked with the PublicApiAnalyzers.

### DIFF
--- a/src/CodeFormatter.cs
+++ b/src/CodeFormatter.cs
@@ -89,6 +89,10 @@ namespace Microsoft.CodeAnalysis.Tools
             foreach (var documentId in documentIdsWithErrors)
             {
                 var documentWithError = solution.GetDocument(documentId);
+                if (documentWithError is null)
+                {
+                    documentWithError = await solution.GetSourceGeneratedDocumentAsync(documentId, cancellationToken).ConfigureAwait(false);
+                }
 
                 logger.LogInformation(Resources.Formatted_code_file_0, documentWithError!.FilePath);
             }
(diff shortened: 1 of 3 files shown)


src/CodeFormatter.cs
```diff
@@ -176,6 +180,7 @@ await DotNetHelper.PerformRestoreAsync(solutionOrProjectPath, logger) != 0)
 
             var documentsCoveredByEditorConfig = ImmutableArray.CreateBuilder<DocumentId>(totalFileCount);
             var documentsNotCoveredByEditorConfig = ImmutableArray.CreateBuilder<DocumentId>(totalFileCount);
+            var sourceGeneratedDocuments = ImmutableArray.CreateBuilder<DocumentId>(16);
```
jasonmalinowski, line 183: Was 16 chosen for any specific reason? You may also consider using a named argument here.

src/CodeFormatter.cs
```diff
@@ -247,6 +258,16 @@ await GeneratedCodeUtilities.IsGeneratedCodeAsync(syntaxTree, cancellationToken)
...
                     }
                 }
+
+                if (formatOptions.IncludeGeneratedFiles)
+                {
+                    var generatedDocuments = await project.GetSourceGeneratedDocumentsAsync(cancellationToken).ConfigureAwait(false);
+                    foreach (var generatedDocument in generatedDocuments)
+                    {
+                        Debug.WriteLine($"Including source generated file '{generatedDocument.FilePath}'.");
+                        sourceGeneratedDocuments.Add(generatedDocument.Id);
```
jasonmalinowski, line 268: Rather than passing IDs around, is it easier to just pass the Document objects around directly?

jasonmalinowski (approved):
Seems fine I guess (I don't know the codebase well enough here) but I do wonder that there's a fair bit of code that's taking lists of DocumentIds and then fetching information about documents, and whether it's better off being refactored a bit to just pass around the Document objects directly.

PR #1486, opened 2022-01-05
Remove unnecessary config

WeihanLi:
This config works only for .NET 6 preview 7, since .NET 6 RC 1 `ImplicitUsings` is used and disabled by default


JoeRobich (approved):
Thanks!

PR #1537, opened 2022-03-02
Simplify logging

### LINKED ISSUE #1536: Simplify the logger creating
Currently, we create the logger via a dependency injection now, think we could create a logger without di to reduce the complexity.
Is there some reason why we get the logger from the dependency injection?

https://github.com/dotnet/format/blob/3f3a6d862960df56a7fae32f41e9c019930080c5/src/Commands/FormatCommandCommon.cs#L115-L125

WeihanLi:
fixes #1536


JoeRobich (approved):
Thanks @WeihanLi! I believe we thought initially that there might use some form of discovery for our Formatters, but we didn't go in that direction and never revisited our usage of DI.

PR #1544, opened 2022-03-08
Add `--exclude-diagnostics` option

JoeRobich:


RikkiGibson (approved):
Do we skip running analyzers whose diagnostics are excluded? It doesn't seem strictly necessary, but was curious.

JoeRobich:
> Do we skip running analyzers whose diagnostics are excluded? It doesn't seem strictly necessary, but was curious.

This is done in the `FilterAnalyzersAsync()` method where we build a map of ProjectId to list of applicable DiagnosticAnalyzers.

https://github.com/dotnet/format/blob/987edc9c5784a7a0b760f1f49e818a74ffdaf463/src/Analyzers/AnalyzerFormatter.cs#L292-L305

jmarolf (approved):

PR #1727, opened 2022-09-22
[release/7.x] Use correct version props for Microsoft.Extensions.* packages

lbussell:
These packages aren't guaranteed to version together, especially in servicing. They should use their own individual properties.

See https://github.com/dotnet/installer/pull/14566 for more context.


JoeRobich (approved):

lbussell:
This is stale/no longer relevant.

PR #1772, opened 2022-12-01
Update Arcade and .NET version in global.json

phil-allen-msft:


global.json
```diff
@@ -1,18 +1,18 @@
...
+    "dotnet": "7.0.100",
     "runtimes": {
       "dotnet": [
         "6.0.3"
       ]
     }
   },
   "msbuild-sdks": {
-    "Microsoft.DotNet.Arcade.Sdk": "8.0.0-beta.22466.3"
+    "Microsoft.DotNet.Arcade.Sdk": "8.0.0-beta.22524.5"
```
jmarolf, line 16: This does not look correct. `Version.Details.xml` also needs to be updated. see https://github.com/dotnet/format/pull/1725. I'll need to check that automatic arcade updates are still working for this repo.

PR #1804, opened 2023-02-28
Unit test fixes for Windows

seriouz:
I've made some changes to the unit tests, so that they are all working with windows.
Almost all changes regarded the end-of-line problem between Windows and Linux/Mac.

### DIFF
--- a/tests/Analyzers/ThirdPartyAnalyzerFormatterTests.cs
+++ b/tests/Analyzers/ThirdPartyAnalyzerFormatterTests.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft.  All Rights Reserved.  Licensed under the MIT license.  See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
@@ -86,21 +87,26 @@ class C
 
         int count = 5;
 
+
+        count = 6;
+
     }
 
 }
 ";
 
-            var expectedCode = @"
+            // Hack for Windows: StyleCop uses the Environment NewLine for its replacement,
+            // so we have take care of it
+            var expectedCode = @$"
 class C
-{
+{{
     void M()
-    {
+    {{
         object obj = new object();
-
-        int count = 5;
-    }
-}
+{Environment.NewLine}        int count = 5;
+{Environment.NewLine}        count = 6;
+    }}
+}}
 ";
 
             var editorConfig = new Dictionary<string, string>()
@@ -163,6 +169,9 @@ class C
 
                 // Prefer using. IDISP017
                 ["dotnet_diagnostic.IDISP017.severity"] = "error",
+
+                // We expect new line as end_of_line in this test
+                ["end_of_line"] = EndOfLineFormatter.GetEndOfLineOption("\n"),
             };
 
             await AssertCodeChangedAsync(testCode, expectedCode, editorConfig, fixCategory: FixCategory.Analyzers, analyzerReferences: analyzerReferences);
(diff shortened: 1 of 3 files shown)


seriouz:
@dotnet-policy-service agree

RikkiGibson:
Windows tests runs are timing out here.

Also seeing a warning in the test project: `/home/vsts/work/1/s/tests/Formatters/OrganizeImportsFormatterTests.cs(3,1): warning IDE0005: Using directive is unnecessary. [/home/vsts/work/1/s/tests/dotnet-format.UnitTests.csproj]`

JoeRobich:
Not sure why these changes would be necessary.

sharwell (changes requested):
In general this isn't the approach we'd want to go with for normalizing the tests. I'll need to review the variety of changes individually to discuss the resolution plan.

tests/Analyzers/ThirdPartyAnalyzerFormatterTests.cs
```diff
@@ -86,21 +87,26 @@ void M()
 
         int count = 5;
 
+
+        count = 6;
```
sharwell, line 91: ❔ Why was this added?

tests/Analyzers/ThirdPartyAnalyzerFormatterTests.cs
```diff
@@ -163,6 +169,9 @@ void M()
 
                 // Prefer using. IDISP017
                 ["dotnet_diagnostic.IDISP017.severity"] = "error",
+
+                // We expect new line as end_of_line in this test
+                ["end_of_line"] = EndOfLineFormatter.GetEndOfLineOption("\n"),
```
sharwell, line 174: ❔ Why was this added?

tests/Analyzers/ThirdPartyAnalyzerFormatterTests.cs
```diff
@@ -210,6 +219,9 @@ void M()
...
```
sharwell, line 224: ❔ Why was this added? This test has `expectedCode` that relies on the environment newline, and the default `end_of_line` value would be the same. This test wouldn't have been failing before this change, provided the user on Windows checked out the code with the _required_ `core.autocrlf=true` configuration.

tests/CodeFormatterTests.cs
```diff
@@ -587,9 +587,13 @@ public async Task GeneratorSolution_NoDiagnosticsReported_WhenNotIncludingGenera
...
-            const string ExpectedPublicApi = @"Greeter
-Greeter.Greet() -> void
-Greeter.Greeter() -> void";
+            var sb = new System.Text.StringBuilder();
+            sb.Append("Greeter");
+            sb.Append(Environment.NewLine);
+            sb.Append("Greeter.Greet() -> void");
+            sb.Append(Environment.NewLine);
+            sb.Append("Greeter.Greeter() -> void");
+            var expectedPublicApi = sb.ToString();
```
sharwell, line 596: ❔ Why was this changed?

tests/Formatters/OrganizeImportsFormatterTests.cs
```diff
@@ -41,7 +41,7 @@ class C
 
             var editorConfig = new Dictionary<string, string>()
             {
-                ["end_of_line"] = EndOfLineFormatter.GetEndOfLineOption(Environment.NewLine),
+                ["end_of_line"] = EndOfLineFormatter.GetEndOfLineOption("\n"),
```
sharwell, line 44: ❔ Why was this changed? Applies to all instances of this in the file.

sharwell (commented):
I suspect the author of this pull request might have not set `core.autocrlf=true` on a Windows machine, which would lead to expected failures. Prior to any fixes, please verify that the local configuration is correct.

JoeRobich:
Closing as 9.x development has moved into the dotnet/sdk repo.

PR #1812, opened 2023-04-20
Remove unused dependencies and clean-up & Update dependencies from roslyn and arcade

ViktorHofer:
1. Clean-up package dependencies that aren't needed.
2. The "<?xml" tags aren't required anymore by msbuild
3. Define versions in Versions.props consistently. Consider later adding
   dependabot support for external dependencies.
4. Move common properties into Directory.Build.props (ie LangVersion).
5. Remove unused dotnet7 feed.
6. Use reference assemblies package used to compile the tests for
    running the tests.
7. Update dependencies from arcade and roslyn
8. Update integration test repo SHAs to be able to build those repos
    with the now updated repo local 8.0 P3 SDK.

This fixes the observed issues in the "Code velocity chat" and cleans up the repository.

### Version changes
Commit 1 and 2 upgrade dependencies (Arcade + Roslyn)

Commit 3 updates the following versions:
MicrosoftBuildVersion -> 17.3.2
Microsoft.Extensions.FileSystemGlobbing -> 7.0.0
Microsoft.Extensions.Logging -> 7.0.0
BenchmarkDotNet -> 0.13.5
BenchmarkDotNet.Annotations -> 0.13.5

Microsoft.DiaSymReader -> 2.0.0-beta-22503-02 (transitive dependency)

Microsoft.VisualBasic -> 10.3.0 (transitive test dependency)
NETStandard.Library -> 2.0.3 (transitive test dependency)
NuGetVersion -> 6.6.0-preview.3.61 (transitive test dependency)
Microsoft.VisualStudio.Composition -> 17.4.16 (transitive test dependency)


ViktorHofer:
Questions for @mmitche and @MichaelSimons: Should dotnet/format add a subscription to roslyn-analyzers and runtime so that dependencies keep updated?

We could also add a toolset dependency for nuget which is currently brought in just for test purposes.

eng/Versions.props
```diff
@@ -1,33 +1,48 @@
...
     <SystemCommandLineRenderingVersion>0.4.0-alpha.22564.1</SystemCommandLineRenderingVersion>
+    <!-- corefx -->
+    <MicrosoftVisualBasicVersion>10.3.0</MicrosoftVisualBasicVersion>
+    <!-- msbuild -->
+    <MicrosoftBuildVersion>17.3.2</MicrosoftBuildVersion>
+    <!-- netstandard -->
+    <NETStandardLibraryVersion>2.0.3</NETStandardLibraryVersion>
+    <!-- nuget -->
+    <!-- In order tests against the same version of NuGet as the SDK. We have to set this to match. -->
+    <NuGetVersion>6.6.0-preview.3.61</NuGetVersion>
```
ViktorHofer, line 31: This is the concerning part in this repository. The tests currently require the version of NuGet assemblies referenced (brought in by  `Microsoft.CodeAnalysis.Analyzer.Testing`) and the local SDK to match. We should try to remove this dependency somehow as a follow-up. Otherwise we would continue to put a high pressure on SBRP. cc @mmitche

azure-pipelines-integration.yml
```diff
@@ -29,43 +29,53 @@ jobs:
...
-        _sha: "b670821b5150bc7ddb01f415d91c17dd55c752a2"
-      sdk:
-        _repo: "https://github.com/dotnet/sdk"
-        _repoName: "dotnet/sdk"
-        _targetSolution: "sdk.sln"
-        _branchName: "main"
-        _sha: "d32c8e7d19f3fc13190ff14cd960370a5d148995"
+        _sha: "a6a78dd64fc1eb466113ece272747a72d497aad7"
+
+      # TODO: Re-enable when formatting issue are investigated.
```
ViktorHofer, line 34: Opened an issue for the disabled integration tests: https://github.com/dotnet/format/issues/1813

eng/format-verifier.ps1
```diff
@@ -73,7 +73,7 @@ try {
 
             if ($stage -eq "format-workspace") {
                 Write-Output "$(Get-Date) - $solutionFile - Formatting Workspace"
-                $output = dotnet.exe "$currentLocation/artifacts/bin/dotnet-format/Release/net7.0/dotnet-format.dll" $solution --no-restore -v diag --verify-no-changes | Out-String
+                $output = dotnet.exe "$currentLocation/artifacts/bin/dotnet-format/Release/net8.0/dotnet-format.dll" $solution --no-restore -v diag --verify-no-changes | Out-String
```
ViktorHofer, line 76: Opened an issue for the hardcoded TFM values: https://github.com/dotnet/format/issues/1814

Directory.Packages.props
```diff
@@ -1,32 +1,30 @@
 <Project>
-  <!-- <Import Project="eng\Versions.props" /> -->
 
   <PropertyGroup>
-    <!-- In order tests against the same version of NuGet as the SDK. We have to set this to match. -->
-    <NuGetVersion>6.4.0</NuGetVersion>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
```
sharwell, line 4: ❔ Is this file always included, or only included when `ManagePackageVersionsCentrally` is set? Just want to make sure it's still applied.

sharwell: ➡️ It's always imported.
https://github.com/NuGet/NuGet.Client/blob/476dca9762230d3212162b4be3f8d77151678711/src/NuGet.Core/NuGet.Build.Tasks/NuGet.props#L35-L39

ViktorHofer: Yes. `ManagePackageVersionsCentrally` is usually set directly in Directory.Packages.props and components that don't want to auto-import that file are expected to set `ImportDirectoryPackagesProps=false`.

eng/Versions.props
```diff
@@ -1,33 +1,48 @@
...
+    <MicrosoftNetCompilersToolsetPackageVersion>4.5.0-3.22602.7</MicrosoftNetCompilersToolsetPackageVersion>
+    <!-- roslyn-sdk -->
+    <MicrosoftCodeAnalysisAnalzerTestingVersion>1.1.2-beta1.22216.1</MicrosoftCodeAnalysisAnalzerTestingVersion>
+    <!-- runtime -->
+    <MicrosoftExtensionsFileSystemGlobbingVersion>7.0.0</MicrosoftExtensionsFileSystemGlobbingVersion>
+    <MicrosoftExtensionsLoggingVersion>7.0.0</MicrosoftExtensionsLoggingVersion>
+    <!-- symreader -->
+    <MicrosoftDiaSymReaderVersion>2.0.0-beta-22503-02</MicrosoftDiaSymReaderVersion>
+    <!-- external -->
+    <BenchmarkDotNetVersion>0.13.5</BenchmarkDotNetVersion>
```
sharwell, line 42: 💭 The current PR is making it difficult to review the _changes_ to versions separate from everything else.

Directory.Build.props
```diff
@@ -1,21 +1,13 @@
...
 
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
-  <!-- Nuget settings -->
   <PropertyGroup>
-    <!-- Use Directory.Packages.props for nuget package versions -->
-    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <Copyright>$(CopyrightMicrosoft)</Copyright>
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <LangVersion>latest</LangVersion>
```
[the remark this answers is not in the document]
ViktorHofer: Just curious, why? We use latest or preview in many repositories in the stack.

tests/Analyzers/AnalyzerAssemblyGenerator.cs
```diff
@@ -30,9 +37,19 @@ private static async Task<IEnumerable<MetadataReference>> GetReferencesAsync()
                 MetadataReference.CreateFromFile(typeof(CodeFixProvider).Assembly.Location),
             };
```
sharwell, line 38: 📝 All of these should be removed and resolved from NuGet packages instead of from assemblies on disk.

Directory.Build.props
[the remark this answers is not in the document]
mmitche: I also prefer latest. Because of its size and relative vijsibility, this repo should be designed to avoid stagnating as much as possible.

eng/Versions.props
ViktorHofer: Agreed. I updated the PR summary to include the version changes.

azure-pipelines-integration.yml
```diff
@@ -29,43 +29,53 @@ jobs:
...
         _targetSolution: "Compilers.sln"
         _branchName: "main"
-        _sha: "b670821b5150bc7ddb01f415d91c17dd55c752a2"
-      sdk:
-        _repo: "https://github.com/dotnet/sdk"
-        _repoName: "dotnet/sdk"
-        _targetSolution: "sdk.sln"
-        _branchName: "main"
-        _sha: "d32c8e7d19f3fc13190ff14cd960370a5d148995"
+        _sha: "a6a78dd64fc1eb466113ece272747a72d497aad7"
```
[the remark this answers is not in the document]
ViktorHofer: Yes - the local SDK is used to build these repositories. The SDK is upgraded to 8.0 P3 with this change which requires the corresponding repositories SHAs to be also updated.

tests/Analyzers/AnalyzerAssemblyGenerator.cs
ViktorHofer: If you don't mind, let's do that in a follow-up as this is already existent in main. Would you mind filing an issue?

Directory.Build.props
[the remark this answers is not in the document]
sharwell: In this case let's just use the most recent fixed version supported by the compiler. I believe that's 11 here, but it might be 10. Feel free to follow-up offline.

azure-pipelines-integration.yml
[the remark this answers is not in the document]
sharwell: Unless this is the last time we're going to update the SDK, this is going to cause a significant stabilization problem for the performance tests. Is there a reason the current SDK can't load the older Roslyn source?

Directory.Build.props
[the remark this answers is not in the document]
ViktorHofer: Updated.

azure-pipelines-integration.yml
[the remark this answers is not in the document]
ViktorHofer: Yes. Repositories like aspnetcore ping the required SDK version without allowing to roll forward: https://github.com/dotnet/aspnetcore/blob/e832fef051ab1256d6d50fc79d1582451f908453/global.json#L3

That means that you can only build the repository with that exact SDK.

azure-pipelines-integration.yml
```diff
@@ -85,4 +75,4 @@ jobs:
       displayName: Run dotnet-format on $(_repoName) $(_targetSolution)
 
     - script: eng\integration-test.cmd -repo '$(_repo)' -branchName '$(_branchName)' -sha '$(_sha)' -targetSolution '$(_targetSolution)' -testPath '$(Agent.TempDirectory)\temp' -stage 'format-folder'
-      displayName: Run dotnet-format on $(_repoName) repo folder
+      displayName: Run dotnet-format on $(_repoName) repo folder
```
sharwell, line 78: This file is missing the trailing newline

tests/Analyzers/AnalyzerAssemblyGenerator.cs
```diff
@@ -30,9 +37,19 @@ private static async Task<IEnumerable<MetadataReference>> GetReferencesAsync()
...
+            var netCurrentTargetingPackVersion = (string)AppContext.GetData("ReferenceAssemblies.NetCurrent.TargetingPackVersion")!;
+            var netCurrentTargetFramework = (string)AppContext.GetData("ReferenceAssemblies.NetCurrent.TargetFramework")!;
+            var nugetConfigPath = (string)AppContext.GetData("ReferenceAssemblies.NetCurrent.NuGetConfigPath")!;
+            ReferenceAssemblies netCurrentReferenceAssemblies = new(netCurrentTargetFramework,
+                new PackageIdentity("Microsoft.NETCore.App.Ref", netCurrentTargetingPackVersion),
+                Path.Combine("ref", netCurrentTargetFramework));
+            netCurrentReferenceAssemblies = netCurrentReferenceAssemblies.WithNuGetConfigFilePath(nugetConfigPath);
+
+            var netcoreMetadataReferences = await netCurrentReferenceAssemblies.ResolveAsync(LanguageNames.CSharp, CancellationToken.None);
             references.AddRange(netcoreMetadataReferences.Where(reference => Path.GetFileName(reference.Display) != "System.Collections.Immutable.dll"));
```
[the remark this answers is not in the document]
ViktorHofer: That code is still valid as System.Collections.Immutable.dll is part of the Microsoft.NETCore.App.Ref targeting pack. I have no context here but I assume the assembly is filtered out as it's coming in via an external dependency and this is a lazy attempt to replicate msbuild's conflict resolution.

This should probably be removed together with your above feedback
> All of these should be removed and resolved from NuGet packages instead of from assemblies on disk. The only time MetadataReference.CreateFromFile(typeof(T).Assembly.Location) is valid is when the test assembly is defined in the same repository as T.

azure-pipelines-integration.yml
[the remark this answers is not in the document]
ViktorHofer: I found a way to use the child repo's SDK instead for restore. That's the best option as it avoids dependencies in dotnet/format itself and uses what the child repo defines. I wonder why this wasn't done to begin with.

sharwell (approved):

PR #1821, opened 2023-04-25
Enable source-build prebuilt detection

MichaelSimons:
Fixes https://github.com/dotnet/format/issues/1809

There are some prebuilts remaining as indicated in SourceBuildrebuiltBaseline.xml that still need addressing.


MichaelSimons:
cc @dotnet/source-build-internal, @mmitche, @oleksandr-didyk

ViktorHofer:
> I think Microsoft.NET.Compilers.Toolset could be unified with the rest of the roslyn dependencies and that would eliminate a huge swath.

Do we need the Microsoft.NET.Compilers.Toolset dependency in this repo at all? It might have been needed in the past but I wonder if we can these days just use the compiler that comes via the toolset.

mmitche:
> > I think Microsoft.NET.Compilers.Toolset could be unified with the rest of the roslyn dependencies and that would eliminate a huge swath.
> 
> Do we need the Microsoft.NET.Compilers.Toolset dependency in this repo at all? It might have been needed in the past but I wonder if we can these days just use the compiler that comes via the toolset.

Probably not, if the repo has UseToolMicrosoftNetCompilersToolset set to false/unset

MichaelSimons:
> Probably not, if the repo has UseToolMicrosoftNetCompilersToolset set to false/unset

UseToolMicrosoftNetCompilersToolset does not appear to be set.

MichaelSimons:
> I think Microsoft.NET.Compilers.Toolset could be unified with the rest of the roslyn dependencies and that would eliminate a huge swath.

The Microsoft.NET.Compilers.Toolset dependency is used to control all roslyn dependencies.  Unfortunately the use of MicrosoftNetCompilersToolsetPackageVersion caused the repo to encounter https://github.com/dotnet/source-build/issues/3392 which caused the repo to build with the MicrosoftNetCompilersToolsetPackageVersion coming from the SDK instead of the repo's versions.props.

MichaelSimons:
@ViktorHofer and @mmitche - I has resolved all prebuilts except one that I propose leaving as an exception for now.  Please review.  TIA

eng/Versions.props
```diff
@@ -29,7 +29,14 @@
...
     <!-- roslyn -->
-    <MicrosoftNetCompilersToolsetPackageVersion>4.7.0-1.23226.15</MicrosoftNetCompilersToolsetPackageVersion>
+    <MicrosoftCodeAnalysisCSharpFeaturesVersion>4.7.0-1.23226.15</MicrosoftCodeAnalysisCSharpFeaturesVersion>
+    <MicrosoftCodeAnalysisCSharpVersion>4.7.0-1.23226.15</MicrosoftCodeAnalysisCSharpVersion>
+    <MicrosoftCodeAnalysisFeaturesVersion>4.7.0-1.23226.15</MicrosoftCodeAnalysisFeaturesVersion>
+    <MicrosoftCodeAnalysisVisualBasicFeaturesVersion>4.7.0-1.23226.15</MicrosoftCodeAnalysisVisualBasicFeaturesVersion>
+    <MicrosoftCodeAnalysisVisualBasicWorkspacesVersion>4.7.0-1.23226.15</MicrosoftCodeAnalysisVisualBasicWorkspacesVersion>
+    <MicrosoftCodeAnalysisVisualBasicVersion>4.7.0-1.23226.15</MicrosoftCodeAnalysisVisualBasicVersion>
+    <MicrosoftCodeAnalysisWorkspacesMSBuildVersion>4.7.0-1.23226.15</MicrosoftCodeAnalysisWorkspacesMSBuildVersion>
+    <MicrosoftCodeAnalysisVersion>4.7.0-1.23226.15</MicrosoftCodeAnalysisVersion>
```
ViktorHofer, line 39: As these are all new versions which presumably aren't used in a non source build build, how do we communicate their intent? Should we add a comment? Will they show up as prebuilt failures if being removed?

ViktorHofer (approved):

eng/Version.Details.xml
```diff
@@ -1,23 +1,79 @@
...
     <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="8.0.0-beta.23253.3">
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>5c98e75aa0078eebd3b8f9d6a6fbed1a1a5eb075</Sha>
+      <SourceBuild RepoName="arcade" ManagedOnly="true" />
+    </Dependency>
+    <Dependency Name="Microsoft.SourceLink.GitHub" Version="8.0.0-beta.23218.3" CoherentParentDependency="Microsoft.DotNet.Arcade.Sdk">
+      <Uri>https://github.com/dotnet/sourcelink</Uri>
+      <Sha>47c52dd2ebf9edfd40abdcff999c13eb461f6ce2</Sha>
+      <SourceBuild RepoName="sourcelink" ManagedOnly="true" />
+    </Dependency>
```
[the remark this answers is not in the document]
MichaelSimons: I must be living under a rock - this is news to me.  If that's the case, then yes all sourcelink dependencies would have to be removed as well as removing the sourcelink repo from the VMR.  

Can you point me to an issue for this work?

eng/Versions.props
MichaelSimons: These are not source-build specific.  Previously the repo was using a single dependency on Microsoft.Net.Compilers.Toolset to represent all of the dependencies from roslyn.  I broken them apart to reference the individual packages.  This felt like the right thing to do based on the questions asked previously about what the Microsoft.Net.Compilers.Toolset dependency was used for.  I don't see a comment as necessary as it is the standard arcade dependency pattern.

MichaelSimons: For clarity, these are all being referenced in https://github.com/dotnet/format/pull/1821/files#diff-5baf5f9e448ad54ab25a091adee0da05d4d228481c9200518fcb1b53a65d4156

MichaelSimons:
cc @sharwell - Can you review and merge if you approve?  TIA.

eng/Versions.props
ViktorHofer: Ah cool, I missed that. Looks good then.

eng/Version.Details.xml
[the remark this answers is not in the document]
ViktorHofer: https://github.com/dotnet/sdk/pull/31632, cc @tmat

mmitche (approved):

eng/Versions.props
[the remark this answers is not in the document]
MichaelSimons: It is not strictly required but it is the arcade pattern.  Updating is trivial when using darc.  That being said, I reverted to a single property.

MichaelSimons:
@sharwell - could you please review and merge if approved?  This PR conflicts with every Arcade dependency flow so it is a burden to keep green.  TIA

sharwell (approved):

PR #1836, opened 2023-05-08
When loading a project, use metadata references when possible.

JoeRobich:
This will not help CI cases where a build has not already been performed. It could significantly speed up project loading on developer machines when they already have a build of referenced projects.

Could help in situations like https://github.com/dotnet/format/issues/1378#issuecomment-1523117354

### DIFF
--- a/src/Workspaces/MSBuildWorkspaceLoader.cs
+++ b/src/Workspaces/MSBuildWorkspaceLoader.cs
@@ -51,6 +51,9 @@ namespace Microsoft.CodeAnalysis.Tools.Workspaces
             {
                 try
                 {
+                    // In the case of loading a specific project, we don't want to load referenced projects, which is very time consuming, when there is already
+                    // a build on disk we can pull metadata from.
+                    workspace.LoadMetadataForReferencedProjects = true;
                     await workspace.OpenProjectAsync(solutionOrProjectPath, msbuildLogger: binlog, cancellationToken: cancellationToken).ConfigureAwait(false);
                 }
                 catch (InvalidOperationException)


src/Workspaces/MSBuildWorkspaceLoader.cs
```diff
@@ -51,6 +51,9 @@ internal static class MSBuildWorkspaceLoader
             {
                 try
                 {
+                    // In the case of loading a specific project, we don't want to load referenced projects, which is very time consuming, when there is already
+                    // a build on disk we can pull metadata from.
+                    workspace.LoadMetadataForReferencedProjects = true;
```
JoeRobich, line 56: I assumed this wouldn't be useful in the solution loading case, since referenced projects would likely be part of the solution. Thinking about it again, it isn't strictly necessary, so this may just need to be enabled generally.

RikkiGibson (approved):
Is this the right base branch for this change?

Any before/after comparison we can do in a project which has already been built on disk?

RikkiGibson:
May need to assign someone to take the PR over if we think this is a useful change (it seems like it probably is?)

zdenek-jelinek:
Hello @RikkiGibson, is there possibly any way to move this forward, please? While the initial comment states that this probably won't help CI cases, I'm more than happy to reorder build/format in my pipeline to make format faster.

I'm not in a hurry to get this but it would be nice to know if any progress is planned or if I should look for other tools instead, the performance of dotnet format in CI is limiting even for smaller projects.

JoeRobich:
Closing as 9.x development has moved into the dotnet/sdk repo.

PR #1850, opened 2023-05-19
Fix format issues in 7.0.3xx SDK

jaredpar:
The repository references the 4.5.0 version of Roslyn and that is what gets loaded into the default `AssemblyLoadContext`. The analyzers in the 7.0.3xx SDK reference the 4.6.0 version of Roslyn. This breaks down
  because there is only one `AssemblyLoadContext` in play here and 4.5.0
  versions are already loaded. Moving to 4.6.0 to unblock the 7.0.3xx
  scenarios.

Related https://github.com/dotnet/sdk/issues/32598

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -14,10 +14,10 @@
   </PropertyGroup>
   <PropertyGroup>
     <MicrosoftBuildVersion>17.3.0-preview-22302-02</MicrosoftBuildVersion>
-    <MicrosoftCodeAnalysisAnalyzersVersion>3.3.3</MicrosoftCodeAnalysisAnalyzersVersion>
+    <MicrosoftCodeAnalysisAnalyzersVersion>3.3.4</MicrosoftCodeAnalysisAnalyzersVersion>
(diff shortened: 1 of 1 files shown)


src/Analyzers/AnalyzerAssemblyLoader.Desktop.cs
```diff
@@ -0,0 +1,100 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#if !NETCOREAPP
```
sharwell, line 5: ❔ This is always excluded right?

sharwell (approved):

jasonmalinowski (approved):

arkalyanms (approved):

JoeRobich:
I created branches with appropriate subscriptions, set channels, and updated SDK subscriptions. https://github.com/dotnet/format/pull/1851

jaredpar:
@JoeRobich should we just close this then and accept your PR as the one to move forward on? It seems like the more standard approach.

JoeRobich:
> @JoeRobich should we just close this then and accept your PR as the one to move forward on? It seems like the more standard approach.

Yup the Roslyn versions should all be updated at this point.

PR #1885, opened 2023-06-26
Fix analyzer and code fix construction

sharwell:
* Replace the custom assembly loader with the one implemented in Roslyn
* Use GetAnalyzers to retrieve analyzers
* Ignore exceptions during extension construction


JoeRobich (approved):
Excited to see this change.

PR #1887, opened 2023-06-27
Add explicit System.Text.Json dependency

NikolaMilosavljevic:
Contributes to: https://github.com/dotnet/source-build/issues/3540

In source-build, `Microsoft.Build` package dependency is pulling a transitive `System.Text.Json` dependency from PSB. The fix is to simply add an explicit `System.Text.Json` dependency.

I'm not 100% sure if changes in `eng/Version.Details.xml` are necessary.

### DIFF
--- a/src/dotnet-format.csproj
+++ b/src/dotnet-format.csproj
@@ -54,6 +54,12 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" />
     <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" />
     <PackageReference Include="Microsoft.Extensions.Logging" />
+
+    <!--
+      Explicit System.Text.Json package reference is required for source-build to pick up the live package.
+      Avoids picking up an old version via transitive dependency from Microsoft.Build or Microsoft.CodeAnalysis.Workspaces.MSBuild.
+    -->
+    <PackageReference Include="System.Text.Json"  Condition=" '$(DotnetBuildFromSource)' == 'true'" />
   </ItemGroup>
 
   <ItemGroup>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -36,5 +36,6 @@
     <PackageVersion Include="NuGet.Versioning" Version="$(NuGetVersion)" />
     <PackageVersion Include="System.CommandLine" Version="$(SystemCommandLineVersion)" />
     <PackageVersion Include="System.CommandLine.Rendering" Version="$(SystemCommandLineRenderingVersion)" />
+    <PackageVersion Include="System.Text.Json" Version="$(SystemTextJsonVersion)" />
   </ItemGroup>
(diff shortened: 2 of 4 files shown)


NikolaMilosavljevic:
@mthalman @MichaelSimons @mmitche

NikolaMilosavljevic:
@mmitche @JoeRobich @sharwell what is the correct branch that would flow into sdk's `6.0.1xx`? I picked `release/6.0.2xx` but there is also the `release/6.x` branch.

JoeRobich:
Whoops, seems like we missed a branch. I've added a `release/6.0.1xx` branch with appropriate subscriptions which publishes to the expected channel. @sharwell I opened https://github.com/dotnet/roslyn-tools/pull/1355 to update the code flow so update flow appropriately.

JoeRobich (approved):

src/dotnet-format.csproj
```diff
@@ -54,6 +54,12 @@
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" />
     <PackageReference Include="Microsoft.Extensions.FileSystemGlobbing" />
     <PackageReference Include="Microsoft.Extensions.Logging" />
+
+    <!--
+      Explicit System.Text.Json package reference is required for source-build to pick up the live package.
+      Avoids picking up an old version via transitive dependency from Microsoft.Build or Microsoft.CodeAnalysis.Workspaces.MSBuild.
+    -->
+    <PackageReference Include="System.Text.Json"  Condition=" '$(DotnetBuildFromSource)' == 'true'" />
```
MichaelSimons, line 62: Is it necessary to make this source-build conditioned?

NikolaMilosavljevic: I wanted to limit the impact of this change, in servicing. Should we try to make it applicable for all scenarios?

MichaelSimons: I'll let the repo owners decide.  My general recommendation is to avoid source-build specific conditions as much as possible.

NikolaMilosavljevic: > I'll let the repo owners decide. My general recommendation is to avoid source-build specific conditions as much as possible.

Thanks. @JoeRobich any concerns with making this change applicable for all scenarios and not just source-build?

@MichaelSimons , I have the same condition in the `sdk` change: https://github.com/dotnet/sdk/pull/33618

NikolaMilosavljevic: @sharwell do you have any concerns with updating this PR to remove the source-build condition and make the new package reference applicable in all build environments?

NikolaMilosavljevic: It seems that we don't have a confirmation yet that source-build condition should be removed. Will merge by EOD unless I hear otherwise.

@sharwell the question was about modifying this PR to remove source-build condition - any thoughts?

PR #1907, opened 2023-07-24
Update MSBuildWorkspaceLoaderTests

sharwell:
Enable and clean up tests

### DIFF
--- a/tests/MSBuild/MSBuildWorkspaceLoaderTests.cs
+++ b/tests/MSBuild/MSBuildWorkspaceLoaderTests.cs
@@ -18,6 +18,13 @@ namespace Microsoft.CodeAnalysis.Tools.Tests.MSBuild
 {
     public class MSBuildWorkspaceLoaderTests
     {
+        // Microsoft.CodeAnalysis.CSharp.ErrorCode
+        private const string ERR_NameNotInContext = "CS0103";
+        private const string ERR_NoEntryPoint = "CS5001";
+
+        // Microsoft.CodeAnalysis.VisualBasic.ERRID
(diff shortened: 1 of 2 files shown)


tests/MSBuild/MSBuildWorkspaceLoaderTests.cs
```diff
@@ -30,26 +37,31 @@ public MSBuildWorkspaceLoaderTests(ITestOutputHelper output)
...
-        // [InlineData("wpfcustomcontrollib")]
-        public async Task CSharpTemplateProject_WindowsOnly_LoadWithNoDiagnostics(string templateName, string ignoredDiagnostics = "")
+        [InlineData("wpf")]
+        [InlineData("wpfusercontrollib")]
+        [InlineData("wpflib")]
+        [InlineData("wpfcustomcontrollib")]
+        public async Task CSharpTemplateProject_WindowsOnly_LoadWithNoDiagnostics(string templateName)
         {
-            await AssertTemplateProjectLoadsCleanlyAsync(templateName, LanguageNames.CSharp, ignoredDiagnostics.Split(";", StringSplitOptions.TrimEntries));
+            var ignoredDiagnostics = templateName switch
```
JoeRobich, line 46: Sad that these diagnostics mean that MSBuildWorkspace isn't able to load those project types correctly.

JoeRobich (approved):

tests/MSBuild/MSBuildWorkspaceLoaderTests.cs
sharwell: I imagine it would be able to if it ran additional build targets to generate the output files for the XAML content.

PR #1933, opened 2023-08-17
Update Readme

JoeRobich:
- Remove link to nuget package. No new releases to NuGet now that we ship in the SDK.
- Fix AzDO build badges. These have been broken since the move to dnceng-public.
- Remove 'New in v5' section. Only a couple releases behind now.


sharwell:
> Remove link to nuget package. No new releases to NuGet now that we ship in the SDK.

Can we add the instructions for installing the global tool with a specific version from the azure artifacts?

JoeRobich:
> Can we add the instructions for installing the global tool with a specific version from the azure artifacts?

Good suggestion. Added a section for it.

sharwell (approved):

Issue #1945, opened 2023-08-23
dotnet-format 8.0 rc1 is failing on Alpine 3.17


NikolaMilosavljevic:
It appears that MSBuildLocator is not finding `hostfxr` library and .NET SDK, thus failing to provide MSBuild path.

Here's the failure output, from source-build tests:

```
System.InvalidOperationException : Failed to execute /src/git/dotnet4/test/Microsoft.DotNet.SourceBuild.SmokeTests/bin/Release/net8.0/.dotnet/dotnet format /src/git/dotnet4/test/Microsoft.DotNet.SourceBuild.SmokeTests/bin/Release/net8.0/projects-202308231813239264/FormatProject/FormatProject.csproj
  Exit code: 3

  Unable to locate MSBuild. Ensure the .NET SDK was installed with the official installer.
```

Perhaps getting rid of MSBuildLocator dependency is good option, considering that `dotnet-format` ships in SDK and should be able to find MSBuild directly.

NikolaMilosavljevic:
I've confirmed that this reproes outside of source-build tests.

I've extracted .NET 8 preview tarball and created symlink for `dotnet` in `/usr/bin`. `dotnet format` command is failing.

```
edb204ed8383:/src/proj8# dotnet new console
The template "Console App" was created successfully.

Processing post-creation actions...
Restoring /src/proj8/proj8.csproj:
  Determining projects to restore...
  Restored /src/proj8/proj8.csproj (in 174 ms).
Restore succeeded.


edb204ed8383:/src/proj8# dotnet format
Unable to locate MSBuild. Ensure the .NET SDK was installed with the official installer.
```

JoeRobich:
https://github.com/microsoft/MSBuildLocator/pull/230

> Perhaps getting rid of MSBuildLocator dependency is good option, considering that dotnet-format ships in SDK and should be able to find MSBuild directly.

I agree. Maybe keep MSBuildLocator as a fallback for those repos who still install as a local tool (see Roslyn).

PR #1948, opened 2023-08-30
Upgrade MSBuildLocator to version 1.6.1 

JoeRobich:
This includes several fixes around resolving hostfxr. (See https://github.com/microsoft/MSBuildLocator/pull/230 & https://github.com/microsoft/MSBuildLocator/pull/236)

resolves https://github.com/dotnet/format/issues/1945


mthalman:
I updated the version of source-build-externals because the previous version contained 1.5.5 of MSBuildLocator. So this update to 1.6.1 cause a source-build prebuilt error to occur. The latest version of source-build-externals has version 1.6.1 so this should resolve the error.

mthalman:
@JoeRobich or @sharwell, could you investigate the build errors? This change is needed to unblock code flow for the sdk: https://github.com/dotnet/installer/pull/17319.

mthalman:
Looks like there are still failures for aspnetcore and efcore.

azure-pipelines-integration.yml
```diff
@@ -23,49 +23,49 @@ jobs:
         _repoName: "dotnet/format"
         _targetSolution: "format.sln"
         _branchName: "main"
-        _sha: "1ec992c6038af9d42499d3ac88c3fd65d0c2f6ed"
+        _sha: "452f6263277533da0af2b557003256d98fbecc27"
```
sharwell, line 26: ❔ Why update all of these?

JoeRobich: Habit. Although I see these repos are now on some very *preview* sdk versions

mthalman (approved):

azure-pipelines-integration.yml
sharwell: If updating the dependency broke formatting for the previous SHA, doesn't that mean some users could see dotnet-format break for the same change?

mthalman:
@dotnet-policy-service rerun

mthalman:
@mmitche - Can you help with this CLA issue?

mmitche:
Nope. But I can merge around it!

PR #1959, opened 2023-09-26
Update Roslyn to 4.8.0-3.23474.1

JoeRobich:
Update Roslyn to a preview 3 build to match the forthcoming .NET 8 RC2 SDK.

### DIFF
--- a/eng/SourceBuildPrebuiltBaseline.xml
+++ b/eng/SourceBuildPrebuiltBaseline.xml
@@ -1,5 +1,5 @@
 <!-- Whenever altering this or other Source Build files, please include @dotnet/source-build-internal as a reviewer. -->
-<!-- See aka.ms/dotnet/prebuilts for guidance on what pre-builts are and how to eliminate them. -->
+<!-- See https://aka.ms/dotnet/prebuilts for guidance on what pre-builts are and how to eliminate them. -->
 
 <UsageData>
   <IgnorePatterns>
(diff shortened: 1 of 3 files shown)


arkalyanms (approved):

arkalyanms:
@Cosifne can we add this to the Tiger checklist if it is not already. Usually dotnet format flies under the Tiger radar.

Cosifne:
Is that possible to let arcade handle the version update? @JoeRobich

Cosifne:
> ~Is that possible to let arcade handle the version update? @JoeRobich~
Nevermind I see the linked arcade PR

Cosifne:
Looks like Roslyn Workspace is introducing a new dependencies:
image
image

@JoeRobich  do you want me to add this to https://github.com/dotnet/format/blob/main/eng/SourceBuildPrebuiltBaseline.xml ?
Or is there any convention to solve this?

Cosifne:
> @Cosifne can we add this to the Tiger checklist if it is not already. Usually dotnet format flies under the Tiger radar.

We can sync at the standup to clarify which repo should be taken care by Tiger. I think it should be part of the arcade flow work. (which this is something unusual)

JoeRobich:
@Cosifne I am not quite sure of the proper fix. @mthalman can you point us in the right direction?

mthalman:
Yes, this can be added to the prebuilt baseline. This is allowed in this case because the prebuilts are limited to this repo and introduced as a dependency of M.CA. That version of M.CA gets overridden by the full product source-build in the VMR. Since it's overridden there, the prebuilts don't occur. They only apply to this repo.

JoeRobich:
Thanks @mthalman! Updated the baseline.

mthalman (approved):

PR #1975, opened 2023-10-19
Fix source build poison leak of Microsoft.Bcl.AsyncInterfaces

mthalman:
Related to https://github.com/dotnet/source-build/issues/3670

This fixes a source build poison leak of Microsoft.Bcl.AsyncInterfaces that comes from this repo. This means the 7.0.0 version of Microsoft.Bcl.AsyncInterfaces defined in https://github.com/dotnet/source-build-reference-packages (SBRP) is leaking into the SDK that is produced by source build. That is a reference assembly and it shouldn't be leaked into the SDK output.

This is caused by a transitive dependency from Microsoft.CodeAnalysis.Workspaces.MSBuild.4.8.0-3.23510.8 to Microsoft.Bcl.AsyncInterfaces.7.0.0. So when loading Microsoft.CodeAnalysis.Workspaces.MSBuild, it also ends up loading Microsoft.Bcl.AsyncInterfaces from the SBRP nuget path.

To fix this, I've added an explicit reference to Microsoft.Bcl.AsyncInterfaces to ensure it uses the latest version, overriding the older version from the transitive dependency.

### DIFF
--- a/src/dotnet-format.csproj
+++ b/src/dotnet-format.csproj
@@ -45,10 +45,6 @@
 
     <PackageReference Include="System.CommandLine" />
     <PackageReference Include="System.CommandLine.Rendering" />
-
-    <!-- Bump the transitive DiaSymReader/1.4.0 dependency which is brought in by the ".Features" packages.
-         Using that verion is undesirable as it brings in the entire .NET Standard 1.x dependency graph. -->
-    <PackageReference Include="Microsoft.DiaSymReader" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/dotnet-format.csproj
+++ b/src/dotnet-format.csproj
@@ -26,6 +26,11 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- Transitive dependency of Microsoft.CodeAnalysis.Workspaces.MSBuild. Need to explicitly provide a reference here to ensure
+         we can get the live version from source build rather than the 7.0.0 version from the transitive reference which leads to
+         a poison leak from source build reference packages. -->
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" />
+
     <PackageReference Include="Microsoft.Build" ExcludeAssets="Runtime" PrivateAssets="All" />
     <PackageReference Include="Microsoft.Build.Framework" ExcludeAssets="Runtime" PrivateAssets="All" />
     <PackageReference Include="Microsoft.Build.Locator" />
--- a/src/dotnet-format.csproj
+++ b/src/dotnet-format.csproj
@@ -26,11 +26,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <!-- Transitive dependency of Microsoft.CodeAnalysis.Workspaces.MSBuild. Need to explicitly provide a reference here to ensure
-         we can get the live version from source build rather than the 7.0.0 version from the transitive reference which leads to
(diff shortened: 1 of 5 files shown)


MichaelSimons (approved):

JoeRobich:
Would it make sense to enable transitive version pinning for these types of issues?

src/dotnet-format.csproj
```diff
@@ -26,6 +26,11 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- Transitive dependency of Microsoft.CodeAnalysis.Workspaces.MSBuild. Need to explicitly provide a reference here to ensure
+         we can get the live version from source build rather than the 7.0.0 version from the transitive reference which leads to
+         a poison leak from source build reference packages. -->
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" />
```
ViktorHofer, line 32: I would prefer if we enable transitive pinning in this repository instead of adding this package reference here.

ViktorHofer: If we keep this then this needs a PrivateAssets attribute.

ViktorHofer (commented):
Remove this part as well now: https://github.com/dotnet/format/blob/2f311969142c7d701724d450376410cdc51039d1/src/dotnet-format.csproj#L49-L51

ViktorHofer:
When we enabled transitive pinning in Arcade, we had some fallout, i.e. SBRPs vs N-1 vs N dependencies. Do you have the possibility to test these changes at the consumption point?

mthalman:
> When we enabled transitive pinning in Arcade, we had some fallout, i.e. SBRPs vs N-1 vs N dependencies. Do you have the possibility to test these changes at the consumption point?

Yeah, I'll run a full build on these changes in the VMR.

ViktorHofer:
One more: https://github.com/dotnet/format/blob/bb8884a00e4ea78a81c1acf0d0544e0711cdf8cf/tests/dotnet-format.UnitTests.csproj#L27-L31C70

ViktorHofer (approved):

ViktorHofer:
I'm not sure but I hope these changes get forward ported into main.

JoeRobich:
> I'm not sure but I hope these changes get forward ported into main.

Looks like we will need to update the roslyn-tools merge config. https://github.com/dotnet/roslyn-tools/blob/main/src/GitHubCreateMergePRs/config.xml#L96-L104

mthalman:
VMR pipeline ran clean with these changes. Confirmed that the reference package doesn't show up in the SDK output.

ViktorHofer (approved):

ViktorHofer:
Unfortunately I don't have merge permissions. Someone else needs to approve and hit the button.

NikolaMilosavljevic:
> VMR pipeline ran clean with these changes. Confirmed that the reference package doesn't show up in the SDK output.

Same SBRP binary was also showing in `dotnet-format` nupkg. @mthalman I presume that instance was also fixed - SDK gets dotnet-format content via this nupkg.

mthalman:
> > VMR pipeline ran clean with these changes. Confirmed that the reference package doesn't show up in the SDK output.
> 
> Same SBRP binary was also showing in `dotnet-format` nupkg. @mthalman I presume that instance was also fixed - SDK gets dotnet-format content via this nupkg.

Right. That's what causes it to flow into the SDK.

PR #2111, opened 2024-02-15
Prepare the FabricBot config for migration to Policy Service

jeffhandley:
This updates the FabricBot configuration to address issues we've seen migrating other repositories to the Policy Service automation. There is one behavioral change with a loss of functionality.

1. Remove the auto-merge automation
    - This automation configuration was relying on some rich features of the FabricBot's own auto-merge functionality
    - That functionality was in place before GitHub's own auto-merge functionality
    - Because GitHub has built-in auto-merge capabilities, the Policy Service does not provide its own implementation
    - Policy Service does have automation that allows for a PR to have auto-merge enabled or disabled, but it can't rely on the types of conditions used here
    - The result is there is not a way to port this configuration forward, and a different approach is needed
    - The recommendation is to investigate Arcade-based auto-merge behaviors or implementing a GitHub Action
2. Explicitly add a filter for `isIssue` or `isPr` where that was previously implied
    - There was a task configured for `"eventType" that when ported to Policy Service, this implicit filter will be lost
    - This updates the config to explicitly add that filter so behavior stays the same after the migration
3. Remove references to project automation (it was superfluous)

/cc @mkArtakMSFT @wtgodbe @JohannesLampel

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -3,32 +3,6 @@
   "tasks": [
     {
       "taskType": "trigger",
-      "capabilityId": "AutoMerge",
-      "subCapability": "AutoMerge",
-      "version": "1.0",
-      "id": "JrwlBTuoi",
-      "config": {
-        "mergeType": "merge",
-        "label": "auto-merge",
-        "taskName": "Auto merge PRs",
-        "minMinutesOpen": "10",
-        "deleteBranches": true,
-        "requireAllStatuses": false,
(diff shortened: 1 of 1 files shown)


sharwell (commented):
I have no objection to this change, but I don't know enough about Fabric Bot to understand the correctness of the change. Hopefully I can get all these rules converted to Github Actions in time to make the conversion to policy service unnecessary/irrelevant.

PR #2162, opened 2024-03-29
1ES port for servicing

arkalyanms:


Cosifne:
🤔What should we do the original yml? with https://github.com/dotnet/format/blob/main/azure-pipelines-official.yml 
Does this mean we want to create a different pipeline in dnceng?

azure-pipelines-1es.yml
jjonescz, line 1: Agree we should probably change azure-pipelines-official.yml instead of creating a new one

azure-pipelines-1es.yml
```diff
@@ -0,0 +1,126 @@
...
+          displayName: Setup Private Feeds Credentials
+          inputs:
+            filePath: $(Build.SourcesDirectory)/eng/common/SetupNugetSources.ps1
+            arguments: -ConfigFile $(Build.SourcesDirectory)/NuGet.config -Password $Env:Token
+          env:
+            Token: $(dn-bot-dnceng-artifact-feeds-rw)
+        - script: eng\common\cibuild.cmd -configuration Release -prepareMachine /p:Test=false
+          displayName: Windows Build
+        - task: CodeQL3000Finalize@0
+          displayName: CodeQL Finalize
```
jjonescz, line 113: I think CodeQL Initialize and Finalize steps should be removed, they are now injected automatically from 1ES pipeline.

azure-pipelines-1es.yml
```diff
@@ -0,0 +1,126 @@
...
+variables:
+- group: AzureDevOps-Artifact-Feeds-Pats
+extends:
+  template: azure-pipelines/MicroBuild.1ES.Official.yml@MicroBuildTemplate
+  parameters:
+    sdl:
+      sourceAnalysisPool: NetCore1ESPool-Svc-Internal
+    pool:
+      name: AzurePipelines-EO
+      image: AzurePipelinesWindows2022compliantGPT
```
jjonescz, line 22: I think we should use something like this:
```suggestion
      name: netcore1espool-internal
      image: windows.vs2022preview.amd64
```

arkalyanms:
Forcing merging to get us 1ES compliant, we can iteratively fix directly on main since repo is not used otherwise. Cannot enable pipeline without the port. Cannot validate the port with the enablement.

PR #2267, opened 2025-11-14
Disable AzDO dependabot

mmitche:
Disable AzDO dependabot.


JoeRobich (approved):

mmitche:
Ah this is on the archive branch. I've mirrored this over to AzDO, and set that as the default branch there.

PR #2288, opened 2026-06-09
Bump System.Security.Cryptography.Xml to 8.0.3

JoeRobich:


mwiemer-microsoft (approved):

eng/Versions.props
marcpopMSFT: Can you bump microsoft build to 17.11.48 while you're at it?
```suggestion
    <MicrosoftBuildVersion>17.11.48</MicrosoftBuildVersion>
```

mwiemer-microsoft: (I'm fine with updating or staying the same)

mthalman (approved):

